PR #5, opened 2015-10-13
Homework

Sharknevercries:
First try.... 
review=?@snowmantw, @FlowerHop, @chungya


homework/test/test-list.js
```diff
@@ -2,7 +2,14 @@ describe('Test > ', function() {
...
 
-  it('will test some pure functions', function() {
+  it('My pure function sin suc', function() {
     // Write any pure function assertion here.
+    assert.equal(Math.sin(Math.PI / 2), 1);
   });
+
+  it('My pure function sin failed', function () {
+    // Write any pure function assertion here.
+    assert.equal(Math.sin(Math.PI / 2), 0);
```
snowmantw, line 12: Good. Although a better way is to define it in one manager and then call it here.

homework/list.js
```diff
@@ -1,79 +1,146 @@
...
+    },
+
+    /**
+    * When the first glance(user has'n done any action), 
+    * fire a 'note-open' event to load the first note.
+    *
+    * @this {ListManager}
+    */
+    preloadFirstNote() {
+      if (this._listNoteContent.length !== 0) {
```
snowmantw, line 103: If a function body is wrapped by a conditional statement only with one branch, the `if...else` should be moved outside the callee, and to the caller.

homework/list.js
```diff
@@ -1,79 +1,146 @@
...
+      });
+    },
+
+    /**
+    * When a note open, it will fire an 'note-open' event.
+    *
+    * @this {ListManager}
+    * @param {object} event The information such as event source, which is about event.
+    */
+    onNoteOpen(event) {
```
snowmantw, line 87: It's a convention to prefix `on` to a method with the corresponding event. Therefore, this method should be renamed as `onClick`. It's a hidden issue about convention and expressiveness.

homework/list.js
```diff
@@ -1,79 +1,146 @@
...
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
+          // Add bind xhr to specify or make code more clear. !?
+          if (this.readyState === 4 && this.status === 200) {
+            var listData = this.response;
+            // The flow ends here.
+            resolve(listData);
```
snowmantw, line 68: Good, you have successfully "Promisified" it.

homework/list.js
```diff
@@ -1,79 +1,146 @@
...
+    *
+    * @this {ListManager}
+    */
+    start() {
+      this._wrapper = document.querySelector('#note-list-wrapper');
+      this.fetchList()
+          .then(this.updateList.bind(this))
+          .then(this.drawList.bind(this))
+          .then(this.preloadFirstNote.bind(this))
+          .catch(function (error) {
```
snowmantw, line 44: **Serious issue**: you should catch it, definitely. However, you need to **re-throw it** after the logging. Or the following chain will not be interrupted and may perform some unexpected effects.

homework/list.js
```diff
@@ -1,79 +1,146 @@
...
+            console.log(error);
+          });
+      window.addEventListener('click', this);
```
snowmantw, line 47: It's good to see you bind the `this` as the listener, rather than an anonymous function.

PR #8, opened 2015-10-16
Homework

rockwyc992:
review=?@snowmantw,@FlowerHop,@chungya 

已完成Tier1，不同的point建立個別的commit

使用上課教的Manager module重新編寫部分的程式碼

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,21 +1,18 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+var ContentManager = function() {
+  var _wrapper = null;
+};
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+ContentManager.prototype = {
 
-  function drawNote(note) {
+  resetWrapper() {
+    this._wrapper.innerHTML = '';
+  },
+
+  drawNote(note) {
     var title = note.title;
     var h = document.createElement('h2');
     h.textContent = title;
@@ -27,11 +24,25 @@
       p.textContent = passage;
       buff.appendChild(p);
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+    this._wrapper.appendChild(h);
+    this._wrapper.appendChild(buff);
+  },
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+  handleEvent(event) {
+    switch(event.type) {
+        case 'note-open':
+          var note = event.detail;
+          this.resetWrapper();
+          this.drawNote(note);
+          break;
+    }
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+    window.addEventListener('note-open', this);
+  }
+};
+
+exports.ContentManager = ContentManager;
+})(window);
--- a/homework/list.js
+++ b/homework/list.js
@@ -12,7 +12,7 @@ ListManager.prototype = {
   onNoteOpen(event) {
     if (event.target.classList.contains('note-title')) {
       var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
+      var content = this._listNoteContent[id];
       window.dispatchEvent(new CustomEvent('note-open',
         { detail: content }));
     };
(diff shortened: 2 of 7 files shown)


homework/list.js
```diff
@@ -52,28 +45,48 @@
...
+    return new Promise(function(reslove, reject) {
+      var xhr = new XMLHttpRequest();
+      xhr.open('GET', url, true);
+      xhr.responseType = 'json';
+      xhr.onreadystatechange = (function(e) {
+        // Watch out: we have a mysterious unknown 'this'.
+        if (this.readyState === 4 && this.status === 200) {
+          var listData = this.response;
+          // The flow ends here.
+          reslove(listData);
```
snowmantw: Typo

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
-  }
+  fetchList(url) {
+    return new Promise(function(reslove, reject) {
```
snowmantw: Typo: reslove -> resolve

(Although as an argument name it still works well)

homework/content.js
```diff
@@ -27,11 +24,25 @@
...
+          var note = event.detail;
+          this.resetWrapper();
+          this.drawNote(note);
+          break;
+    }
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+    window.addEventListener('note-open', this);
```
snowmantw, line 43: Good. Binding `this` as `listener` for the API is better than an anonymous function.

homework/list.js
```diff
@@ -1,44 +1,37 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+ListManager.prototype = {
 
-  function onNoteOpen(event) {
+  onNoteOpen(event) {
```
snowmantw, line 12: It's a convention to prefix `on` to an event handler with the corresponding event. So this method should be renamed as `onNoteClicked` in fact.

homework/list.js
```diff
@@ -1,44 +1,37 @@
...
         { detail: content }));
     };
-  }
+  },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+  preloadFirstNote() {
+    if (this._listNoteContent.length !== 0) {
```
snowmantw, line 22: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
-  }
+  fetchList(url) {
```
snowmantw, line 51: Good. You have successfully "Promisified" the method.

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList('./demo-list-notes.json').then((function(response) {
+      this.updateList(response);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this)).catch(function(error){
+      console.error(error);
```
snowmantw, line 85: **Serious issue**: you should **re-throw** the error to make sure the following chains would be interrupted. If you don't do that, the next chain will still perform the actions since the error has already been captured.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
   beforeEach(function() {
+    subject = new TestManager;
   });
 
-  it('will test some pure functions', function() {
+  it('will test some pure functions', function(done) {
     // Write any pure function assertion here.
+    subject.set(3, 5);
+    assert.deepEqual(subject.add(), 3+5);
+    done();
```
snowmantw, line 12: If it's not an asynchronous function, you don't need to use the `done` like this.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
+
   beforeEach(function() {
+    subject = new TestManager;
   });
 
-  it('will test some pure functions', function() {
+  it('will test some pure functions', function(done) {
     // Write any pure function assertion here.
+    subject.set(3, 5);
+    assert.deepEqual(subject.add(), 3+5);
```
snowmantw, line 11: Good. Although you should make sure the file is included correctly when you perform the test.

PR #9, opened 2015-10-16
101502503

cxz91493:
review=?@snowmantw
review=?@chungya
review=?@FlowerHop

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,44 @@
 'use strict';
 
 (function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
+    var Content =function() {
+        this._wrapper = document.querySelector('#note-content-wrapper');
+    }
+    
+    var ContentNote = new Content();
+  
+    Content.prototype.start = function() {   //function start()
+        window.addEventListener('note-open', function(event) {
+            var note = event.detail;
+            this.resetWrapper();
+            this.drawNote(note);
+        }.bind(this));
+    }
+
+    Content.prototype.resetWrapper = function(){   //function resetWrapper()
+        this._wrapper.innerHTML = '';
+    }
+
+    Content.prototype.drawNote = function(note) {   //function drawNote(note)
+        var title = note.title;
+        var h = document.createElement('h2');
+        h.textContent = title;
+        var passages = note.passages;
+        var buff = document.createDocumentFragment();
+        passages.forEach(function(passage) {
+            var p = document.createElement('p');
+            p.classList.add('note-passage');
+            p.textContent = passage;
+            buff.appendChild(p);
+        });
+        this._wrapper.appendChild(h);
+        this._wrapper.appendChild(buff);
+    }
+
+  
+
+    document.addEventListener('DOMContentLoaded', function(event) {
+        ContentNote.start();
     });
-  }
-
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
-
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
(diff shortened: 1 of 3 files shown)


homework/list.js
```diff
@@ -1,79 +1,90 @@
...
+List.prototype.start=function() {
+    
+    this.fetchList()
+        .then(function(data){
+            this.updateList(data);
+        }.bind(this))
+        .then(this.drawList.bind(this))
+        .then(this.preloadFirstNote.bind(this));
+        //.then(this.preloadFirstNote().bind(this));
+        //It is wrong because ()
```
snowmantw, line 25: Yes because binding doesn't mean you need to call it immediately.

homework/list.js
```diff
@@ -1,79 +1,90 @@
...
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
+var ListNote = new List() ;
+
+List.prototype.start=function() {
+    
+    this.fetchList()
+        .then(function(data){
```
snowmantw: nit: Redundent binding

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
-    });
-  }
+    var Content =function() {
+        this._wrapper = document.querySelector('#note-content-wrapper');
+    }
+    
+    var ContentNote = new Content();
+  
+    Content.prototype.start = function() {   //function start()
+        window.addEventListener('note-open', function(event) {
```
snowmantw, line 12: You should bind the `listener` argument with `this`, rather than an anonymous function like this.

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
+        this._wrapper.appendChild(h);
+        this._wrapper.appendChild(buff);
+    }
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+  
+
+    document.addEventListener('DOMContentLoaded', function(event) {
```
snowmantw, line 41: You should stop bootstrapping your program like this. Instead of, put them in a centralized method is better for testing and debugging.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
+    this.fetchList()
+        
+        /*.then(function(data){
+        this.updateList(data);
+        }.bind(this))*/
+
+        .then(this.updateList.bind(this))
+        //.then(f)
+        .then(this.drawList.bind(this))
+        .then(this.preloadFirstNote.bind(this));
```
snowmantw, line 23: Good. You have practiced the `then` in chain. However, you should try how to handle the error with `catch`. And if you do that, please note that the error should be **log AND re-throw**. Since if you only log it, the following chain will still perform some unexpected side-effects and screw your program.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
-      onNoteOpen(event);
-    });
-  }
+        this.onNoteOpen(event);
+    }.bind(this));
+   
+}
 
-  function onNoteOpen(event) {
+List.prototype.onNoteOpen = function(event) {
```
snowmantw, line 33: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

snowmantw, line 33: And it's a convention to prefix `on` to an event handler with the corresponding event. So this method should be renamed as `onNoteClicked` in fact.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
-      // Watch out: we have a mysterious unknown 'this'.
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
+List.prototype.fetchList = function(){//afterFetch) {
```
snowmantw, line 70: Good. You have successfully "Promisified" the method.

homework/test/test-list.js
```diff
@@ -4,5 +4,6 @@ describe('Test > ', function() {
 
   it('will test some pure functions', function() {
     // Write any pure function assertion here.
+      assert.deepEqual(Math.sin(0) , 0);
```
snowmantw, line 7: It's OKAY. Although it could be better if you define a method in one manager with such content, and then call it here.

PR #10, opened 2015-10-16
Tier1(except test)

kakason:
review=?@chungya
review=?@snowmantw

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,53 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+    //this is a constructor
+    var TodoContentManager = function() {
+        this._wrapper = document.querySelector('#note-content-wrapper');
+    }
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+    //define its prototype
+    TodoContentManager.prototype = {
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+        start() {
+            window.addEventListener('note-open',this);
+        },
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+        resetWrapper() {
+            this._wrapper.innerHTML = '';
+        },
+
+        drawNote(note) {
+            var title = note.title;
+            var h = document.createElement('h2');
+            h.textContent = title;
+            var passages = note.passages;
+            var buff = document.createDocumentFragment();
+            passages.forEach(function (passage) {
+                var p = document.createElement('p');
+                p.classList.add('note-passage');
+                p.textContent = passage;
+                buff.appendChild(p);
+            });
+            this._wrapper.appendChild(h);
+            this._wrapper.appendChild(buff);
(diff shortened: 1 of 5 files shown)


kakason:
review=?@chungya
review=?@snowmantw

homework/list.js
```diff
@@ -1,79 +1,98 @@
...
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+        start() {
+             new Promise(function (resolve) {
+                 this.fetchList(function(data){
```
snowmantw: It's better to design `fetchList` as a function return Promise, so there is no need to use callback style.

For example:

```
this.fetchList()
  .then((function() {
    this.updateList(data);
    this.drawList();
    this.preloadFirstNote();
  }).bind(this))
```

It also clear the code since there is no need to wrap another Promise on that.

kakason:
review=?@chungya
review=?@snowmantw

homework/content.js
```diff
@@ -1,37 +1,53 @@
...
+    //this is a constructor
+    var TodoContentManager = function() {
+        this._wrapper = document.querySelector('#note-content-wrapper');
+    }
+
+    //define its prototype
+    TodoContentManager.prototype = {
+
+        start() {
+            window.addEventListener('note-open',this);
```
snowmantw, line 14: It's good to see you bind the `this` as the `listener, rather than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+        start() {
+                 this.fetchList().then(function(data) {
```
snowmantw, line 13: Good. Although you should try to compose the chain with multiple `then` and `catch`. And you should note that the `catch` should **log AND re-throw** the error, to prevent the following chains keep the error in silent and perform some unexpected effects.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
+        },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+        onNoteOpen(event) {
```
snowmantw, line 21: It's a convention to prefix `on` to an event handler with the corresponding event. So this method should be renamed as `onNoteClicked` in fact.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
+            if (event.target.classList.contains('note-title')) {
+                var id = event.target.dataset.noteId;
+                var content = this._listNoteContent[id];
+                window.dispatchEvent(new CustomEvent('note-open',
+                    { detail: content }));
+            }
+        },
+
+        preloadFirstNote() {
+            if (this._listNoteContent.length !== 0) {
```
snowmantw, line 31: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,95 @@
...
-      li.classList.add('note-title');
-      li.textContent = note.title;
-      // Note: buff is captured, so we now have a
-      // little closure naturally.
-      buff.appendChild(li);
-    });
-    ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+        fetchList() {
```
snowmantw, line 60: Good. You have successfully "Promisifed" it.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
-    // Write any pure function assertion here.
-  });
+    beforeEach(function () {
+        subject = new TodoListManager();
+    });
+
+    it('Test drawList function', function () {
+        subject._wrapper = document.createElement('div');
+        subject.drawList();
+        assert.isNull(subject._wrapper.querySelector('li'), "The test fail!");
```
snowmantw, line 11: Good. You tested a real method, rather than a mocked one.

PR #11, opened 2015-10-16
Homework

ylin0603:
Finish Tier1 and add comment
First try

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,68 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
-
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
-
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
-
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+(function(exports) {
+    /**
+     * Initiate wrapper.
+     * @constructor
+     * @this {NoteContentManager}
+     *
+     */
+    var NoteContentManager = function(){
+        this._wrapper=null;    
+    }
+    NoteContentManager.prototype={
+        /**
+         * Add a event listener to listen to note-open event
+         * @this {NoteContentManager}
+         *
+         */
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', this );
+        },
+        /**
+         * Clean the wrapper.
+         * @this {NoteContentManager}
+         */
+        resetWrapper() {
+            this._wrapper.innerHTML = '';
+        },
+        /**
+         * Handle event
+         *  @param{object} event - custom event :'note-open' 
+         *                          , fire when note is opened. 
+         *  @this{NoteContentManager}
+         */
+        handleEvent(event){
+            switch(event.type){
+                case 'note-open':
+                    var note = event.detail;
+                    this.resetWrapper();
+                    this.drawNote(note);
+                    break;
+            }
+        },
+        /**
+         * Draw the note on wrapper.
+         *
+         * @param{object} note- note to be drawed
+         * @this{NoteContentManager}
+         */
+        drawNote(note) {
+            var title = note.title;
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,68 @@
...
+    /**
+     * Initiate wrapper.
+     * @constructor
+     * @this {NoteContentManager}
+     *
+     */
+    var NoteContentManager = function(){
+        this._wrapper=null;    
+    }
+    NoteContentManager.prototype={
```
ylin0603, line 13: Deal with object properties w/ constructor & prototype

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+        },
+        /**
+         * Fetch Data(note-list) from database,where is localhost.
+         * If success,then continue the following flow 
+         * else it will reject
+         * @this {NoteListManager}
+         * @return {Promise}
+         */
+        fetchList() {
+            return new Promise((function(resolve,reject){
```
ylin0603, line 108: Deal with asynchronous flows with Promise

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+                        var listData = this.response;
+                        // The flow ends here.
+                        resolve(listData);
+                    } else if (this.status !== 200 ){
+                        // Ignore error in this case.
+                        reject('Something Wrong');
+                    }
+                };
+                xhr.send();
+            }).bind(this));
```
ylin0603, line 124: Bind this in Event handlers

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+        }, 
+        /**
+         * Test Function , will add num1 to num2 and return.
+         * @param {number,number}
+         * @return {number}
+         */
+        pureFunction(num1,num2){
```
ylin0603, line 131: Test pure function

homework/test/test-list.js
```diff
@@ -1,8 +1,12 @@
-describe('Test > ', function() {
+describe('Test NoteListManager ', function() {
```
ylin0603, line 1: My first test case

homework/content.js
```diff
@@ -1,37 +1,68 @@
...
+        /**
+         * Add a event listener to listen to note-open event
+         * @this {NoteContentManager}
+         *
+         */
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', this );
```
snowmantw, line 21: It's good to see you bound the `this` as `listener` in the argument list, because there are lots of incorrect code put an anonymous function here.

homework/content.js
```diff
@@ -1,37 +1,68 @@
...
+        /**
+         * Clean the wrapper.
+         * @this {NoteContentManager}
+         */
+        resetWrapper() {
+            this._wrapper.innerHTML = '';
+        },
+        /**
+         * Handle event
+         *  @param{object} event - custom event :'note-open'
```
snowmantw, line 32: JSDoc; good.

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+     * after fetch , draw the list
+     * 
+     * @this {NoteListManager}
+     */
+        start() {
+            window.addEventListener('click',this);
+            this._wrapper = document.querySelector('#note-list-wrapper');  
+            this.fetchList()
+            .then((function(data) {
+                this.updateList(data);
```
snowmantw, line 28: Good. Although you could try to compose a chain with multiple `then` and `catch`. And you need to note that the `catch` should **log AND re-throw** the exception, otherwise the following chains may make the error in silent and perform some unexpected effects after this one.

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+                    break;
+            } 
+        },
+        /**
+         * Dispatch a custom event when mouse click at the note-title.
+         * Also determine the content to be showed.
+         * @this {NoteListManager}
+         * @param {Object} event- mouse click event.
+         */
+        onNoteOpen(event) {
```
snowmantw, line 54: It's a convention to prefix `on` to an event handler with the corresponding event. So this method should be renamed as `onNoteClicked` in fact.

homework/list.js
```diff
@@ -1,79 +1,136 @@
...
+            var content = this._listNoteContent[id];
+            window.dispatchEvent(new CustomEvent('note-open',
+                                                 { detail: content }));
+        },
+        /**
+         * Show the Note for start stage.
+         * @this {NoteListManager}
+         */
+        preloadFirstNote() {
+            if (this._listNoteContent.length !== 0) {
```
snowmantw, line 65: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

homework/list.js
snowmantw, line 108: Good. You have successfully "Promisified" that.

homework/test/test-list.js
```diff
@@ -1,8 +1,12 @@
...
+    var subject;
   beforeEach(function() {
+      subject=new NoteListManager();
   });
 
   it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+    var manager = subject;
+    var result=manager.pureFunction(3,4);
+    assert.equal(result,7,'is it equal to 7');
```
snowmantw, line 10: Good. It's correctly what I asked for.

PR #13, opened 2015-10-17
102502515

wbhan95:


homework/content.js
```diff
@@ -1,37 +1,73 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
+	
+	/**
+	 *Creates an instance of TodoContentManager.
+	 *
+	 *@constructor
```
snowmantw, line 8: JSDoc; good.

homework/content.js
```diff
@@ -1,37 +1,73 @@
...
+	 *Handle the event.
+	 *
+	 *@param {object} Only 'note-open' will happen here.
+	 *@this {TodoContentManager}
+	 */
+    TodoContentManager.prototype.handleEvent = function(event) {
+        switch (event.type) {
+            case 'note-open':
+                var note = event.detail;
+                this.resetWrapper.bind(this)();
```
snowmantw, line 56: **WRONG**: if you call `addEventListener` correctly, the `this` is already bound. Please check with that again, since code like this just show you didn't understand what I've taught in the class.

homework/list.js
```diff
@@ -1,79 +1,135 @@
...
+        };
+    };
+	
+	/**
+	 *When the html open, the first/default passage will open.
+	 *
+	 *@this {TodoListManager}
+	 */
+    TodoListManager.prototype.preloadFirstNote = function() {
+        if (this._listNoteContent.length !== 0) {
```
snowmantw, line 61: If a function body is wrapped with one conditional statement with only one branch, the `if...else` should be put in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,135 @@
...
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
+	/**
+	 *While opening a note, here will be done.
+	 *
+	 *@param {object}
+	 *@this {TodoListManager}
+	 */
+    TodoListManager.prototype.onNoteOpen = function(event) {
```
snowmantw, line 45: In fact it's a convention to prefix a handler with `on` to the corresponding event. So it should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -1,79 +1,135 @@
...
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
+	/**
+	 *To get the data(passage) from demo-list-notes.json.
+	 *
+	 *@return {Promise} work flow
+	 *@this {TodoListManager}
+	 */
+    TodoListManager.prototype.fetchList = function(afterFetch) {
```
snowmantw, line 85: Since you have successfully "Promisifed" this method, you don't need the callback anymore.

homework/list.js
```diff
@@ -1,79 +1,135 @@
...
+		
+	/**
+	 *Start here. get the list id.
+	 *
+	 *@this {TodoListManager}
+	 */
+    TodoListManager.prototype.start = function() {
+        this._wrapper = document.querySelector('#note-list-wrapper')
+        this.fetchList()
+            .then((function(data) {
```
snowmantw, line 127: You should practice to compose a chain with multiple `then` and `catch`. And you should note that `catch` must come with `logging AND re-throw`, or the following chains may screw you, if there are any error in the chain.

homework/test/test-list.js
```diff
@@ -1,8 +1,17 @@
...
+	var assert = require("assert");
+	  it('will test sin', function() {
+		assert.equal(Math.sin(0), 0);
+      });
+	  it('test function', function() {
+		var manager = new Manager();
+		assert.equal('2', manager.add(1, 1));
+	  })
+    });
+	var Manager = function() {};
```
snowmantw, line 14: It's OK, although I supposed you to define this method in one manager above.

PR #14, opened 2015-10-19
Homework

star096374:
Finish Tier1

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -7,33 +7,33 @@
 
     ListNoteContentManager.prototype = {
         start() {
-                this._wrapper = document.querySelector('#note-content-wrapper');
-                window.addEventListener('note-open', (function(event) {
-                    var note = event.detail;
-                    this.resetWrapper();
-                    this.drawNote(note);
-                }).bind(this));
-            },
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', (function(event) {
+                var note = event.detail;
+                this.resetWrapper();
+                this.drawNote(note);
+            }).bind(this));
+        },
 
-            resetWrapper() {
-                this._wrapper.innerHTML = '';
-            },
+        resetWrapper() {
+            this._wrapper.innerHTML = '';
+        },
 
-            drawNote(note) {
-                var title = note.title;
-                var h = document.createElement('h2');
-                h.textContent = title;
-                var passages = note.passages;
-                var buff = document.createDocumentFragment();
-                passages.forEach(function(passage) {
-                    var p = document.createElement('p');
-                    p.classList.add('note-passage');
-                    p.textContent = passage;
-                    buff.appendChild(p);
-                });
-                this._wrapper.appendChild(h);
-                this._wrapper.appendChild(buff);
-            }
+        drawNote(note) {
+            var title = note.title;
+            var h = document.createElement('h2');
+            h.textContent = title;
+            var passages = note.passages;
+            var buff = document.createDocumentFragment();
+            passages.forEach(function(passage) {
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,40 @@
...
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+    ListNoteContentManager.prototype = {
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', (function(event) {
```
snowmantw, line 11: You should bind the listener with `this`, rather than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+    ListNoteManager.prototype = {
+        start() {
+            this._wrapper = document.querySelector('#note-list-wrapper');
+            this.fetchList().then((function(data) {
```
snowmantw, line 12: You should try to put it in a chain with multiple `then` and `catch`. Please note a `catch` should capture an error and re-throw it after logging it.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+        onNoteOpen(event) {
+            if (event.target.classList.contains('note-title')) {
```
snowmantw, line 23: In fact it's a convention to prefix a handler with `on` to the corresponding event. So it should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+        preloadFirstNote() {
+            if (this._listNoteContent.length !== 0) {
```
snowmantw, line 33: If a function body is wrapped with one conditional statement with only one branch, the `if...else` should be put in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
+            return new Promise((function(resolve, reject) {
+                var xhr = new XMLHttpRequest();
+                xhr.open('GET', '/demo-list-notes.json', true);
+                xhr.responseType = 'json';
+                xhr.onreadystatechange = function(e) {
+                    // Watch out: we have a mysterious unknown 'this'.
+                    if (this.readyState === 4 && this.status === 200) {
+                        var listData = this.response;
+                        // The flow ends here.
+                        resolve(listData);
```
snowmantw, line 73: Good. You have successfully "Promisifed" it.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
 
-  it('will test some pure functions', function() {
-    // Write any pure function assertion here.
-  });
+    it('Test resetWrapper function', function() {
+        // Write any pure function assertion here.
+        subject._wrapper = document.createElement('div');
+        subject._wrapper.innerHTML = 'Test';
+        subject.resetWrapper();
+        assert.equal(subject._wrapper.innerHTML, '', "Reset Failed!");
```
snowmantw, line 12: Good. You tested a real method rather than a mocked one.

PR #16, opened 2015-10-20
Lesson 03 101502551

bdsword:
我還想再寫Tier2，只是先推上來比較保險，這樣可以嗎？

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,21 +1,24 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
+  var ContentManager = function () {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+  }
+
+  ContentManager.prototype.start = function () {
     window.addEventListener('note-open', (function(event) {
       var note = event.detail;
-      resetWrapper();
-      drawNote(note);
+      this.resetWrapper();
+      this.drawNote(note);
     }).bind(this));
   }
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
+  ContentManager.prototype.resetWrapper = function () {
+    this._wrapper.innerHTML = '';
   }
 
-  function drawNote(note) {
+  ContentManager.prototype.drawNote = function (note) {
     var title = note.title;
     var h = document.createElement('h2');
     h.textContent = title;
@@ -27,11 +30,9 @@
       p.textContent = passage;
       buff.appendChild(p);
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
+    this._wrapper.appendChild(h);
+    this._wrapper.appendChild(buff);
   }
 
-  document.addEventListener('DOMContentLoaded', (function(event) {
-    start();
-  }).bind(this));
-})();
+  exports.ContentManager = ContentManager;
+})(window);
--- a/homework/content.js
+++ b/homework/content.js
@@ -52,12 +52,14 @@
     h.textContent = title;
     var passages = note.passages;
     var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
+
+    for (var i = 0; i < passages.length; i += 1) {
       var p = document.createElement('p');
       p.classList.add('note-passage');
-      p.textContent = passage;
+      p.textContent = passages[i];
       buff.appendChild(p);
-    });
+    }
+
     this._wrapper.appendChild(h);
     this._wrapper.appendChild(buff);
   }
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,11 +1,28 @@
 'use strict';
 
+/**
+ * @typedef {Object} Note
+ * @property {String} title
+ * @property {String} passages
+ */
+
 (function(exports) {
 
+  /**
+   * Create a instance of ContentManager
+   *
+   * @constructor
+   * @this {ContentManager}
+   */
   var ContentManager = function () {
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,68 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+/**
+ * @typedef {Object} Note
```
snowmantw, line 4: JSDoc; good.

homework/content.js
```diff
@@ -1,37 +1,68 @@
...
+    this._wrapper = document.querySelector('#note-content-wrapper');
+  }
+
+  /**
+   * Add an event handler for note-open event which draw correct note contents
+   *
+   * @this {ContentManager}
+   */
+  ContentManager.prototype.start = function () {
+    window.addEventListener('note-open', (function(event) {
```
snowmantw, line 27: You should bind the `listener` with `this`, rather than an anonymous function like this.

homework/list.js
```diff
@@ -1,79 +1,149 @@
...
+   *
+   * @this {ListManager}
+   */
+  ListManager.prototype.start = function(){
+      this.fetchList().then((function(data) {
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this)).catch((function () {
+        // Deal with fetch error
```
snowmantw, line 34: **Serious issue**: you can't just write a chain and discard all the error. In fact, you need to **log AND re-throw** it. To log the error is to prevent the following chain make it silent and then you're in trouble o debugging, and to re-throw it is to make sure the following will be interrupted if there is an error.

homework/content.js
bdsword, line 27: What do you mean bind listener with this?

Do you mean that I should extract the anonymous function as a method of the ContentManager?

So the 'this' would automatically be ContentManager itself?

homework/list.js
```diff
@@ -1,79 +1,149 @@
...
+    return new Promise(function (resolve, reject) {
+      var xhr = new XMLHttpRequest();
+      xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
+      xhr.responseType = 'json';
+      xhr.onreadystatechange = (function(e) {
+        // Watch out: we have a mysterious unknown 'this'.
+        if (xhr.readyState === 4 && xhr.status === 200) {
+          var listData = xhr.response;
+          // The flow ends here.
+          resolve(listData);
```
snowmantw, line 125: Good. You have successfully "Promisified" the method.

homework/test/test-list.js
```diff
@@ -1,8 +1,10 @@
-describe('Test > ', function() {
-  beforeEach(function() {
-  });
+describe('Test ListManager getFahrenheit', function() {
+  var listManager = new ListManager();
 
-  it('will test some pure functions', function() {
+  it('would calculate correct degree in Fahrenheit scale', function() {
     // Write any pure function assertion here.
+    assert.equal(32.000, listManager.getFahrenheit(0));
```
snowmantw, line 6: Good work.

homework/list.js
```diff
@@ -1,79 +1,149 @@
...
+  },
+
+
+  /**
+   * Preload the first note at the first time this instance start
+   *
+   * @this {ListManager}
+   */
+  ListManager.prototype.preloadFirstNote = function() {
+    if (this._listNoteContent.length !== 0) {
```
snowmantw, line 63: If a function body is wrapped with one conditional statement with only one branch, the `if...else` should be put in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,149 @@
...
+      }).bind(this));
+  };
+
+
+  /**
+   * Dispatch the note-open event and prepare the param for the dispatched event
+   *
+   * @this {ListManager}
+   */
+  ListManager.prototype.onNoteOpen = function(){
```
snowmantw, line 47: In fact it's a convention to prefix a handler with `on` to the corresponding event. So it should be renamed as `onNoteClicked`.

homework/list.js
bdsword, line 34: Ok, I think you are absolutely right. Sorry for the issue here, I just don't know how to handle error when I working on this part, 'cause it's not mentioned in the ppt.

homework/list.js
bdsword, line 47: Ok, naming issue... I would correct it.

homework/content.js
snowmantw, line 27: See the second code example in:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#The_value_of_this_within_the_handler

That's what I asked you to bind `this`: via call `addEventListener` with an object instance, its `handleEvent` will be automatically bound with the object you wish as its `this`.

homework/list.js
snowmantw, line 34: My slides apparently couldn't include all the things, or it will be at least 200+ pages and nobody want to see that. 
That's why I asked you to ask me everything you feel confused or unclear. And since clearly in the classroom and these 3 or 4 weeks nobody did that, I suppose you all know the things I have and haven't taught, until you and other individuals asked in private like this, which is acceptable for me.

homework/list.js
snowmantw, line 47: It's a hidden issue because I knew the course requirement is "learned or understood JavaScript". Just like I won't claim that I "knew C++" in front of my other Gecko colleagues, I suppose that a JavaScript programmer should notice convention issue.

homework/list.js
bdsword, line 63: Agree with you.

PR #17, opened 2015-10-20
homework

RajonRubio:
do homework

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,40 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
+    var ContentManager = function() {
+        this._wrapper = null;
+    };
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+    ContentManager.prototype = {
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', (function(event) {
+                var note = event.detail;
+                this.resetWrapper();
+                this.drawNote(note);
+            }).bind(this));
+        },
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+        resetWrapper() {
+            this._wrapper.innerHTML = '';
+        },
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+        drawNote(note) {
+            var title = note.title;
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,40 @@
...
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+    ContentManager.prototype = {
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', (function(event) {
```
snowmantw, line 11: You should set `this` as the listener as the second example in the manual:

homework/list.js
```diff
@@ -1,79 +1,88 @@
...
+    var _listNoteContent = [];
+    var _wrapper = null;
+  };
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  ListNoteManager.prototype = {
+    start() {
+        this._wrapper = document.querySelector('#note-list-wrapper');
+        this.fetchList().then((function(data) {
```
snowmantw, line 12: You should try to compose the promise chain with multiple `then` and `catch`. And you should notice that the `catch` should **log AND re-throw** the exception, to make sure the error is logged in the console and there won't be any exceptions after the error.

homework/list.js
```diff
@@ -1,79 +1,88 @@
...
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+      onNoteOpen(event) {
```
snowmantw, line 22: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onNoteClick'.

homework/list.js
```diff
@@ -1,79 +1,88 @@
...
-
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+      preloadFirstNote() {
+        if (this._listNoteContent.length !== 0) {
```
snowmantw, line 33: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,88 @@
...
-    xhr.onreadystatechange = function(e) {
-      // Watch out: we have a mysterious unknown 'this'.
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
+      fetchList() {
```
snowmantw, line 63: Good. You have successfully "Promisified" it.

homework/test/test-list.js
```diff
@@ -1,8 +1,16 @@
...
   beforeEach(function() {
+    subject = new ContentManager();
   });
 
-  it('will test some pure functions', function() {
+  it('Test resetWrapper', function() {
     // Write any pure function assertion here.
+    subject._wrapper = document.createElement('div');
+    subject._wrapper.innerHTML = "XXXX";
+    subject.resetWrapper();
```
snowmantw, line 12: Good. You tested the real method rather than a mocked one.

PR #20, opened 2015-10-23
102502048

ZenithMeteor:
Bind this in Event handlers
Deal with object properties w/ constructor & prototype
Deal with asynchronous flows with Promise
Write your first test case

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -13,11 +13,11 @@
       }).bind(this));
   }
 
-  function resetWrapper() {
+  con.prototype.function resetWrapper = function() {
     _wrapper.innerHTML = '';
   }
 
-  function drawNote(note) {
+  con.prototype.function drawNote = function(note) {
     var title = note.title;
     var h = document.createElement('h2');
     h.textContent = title;
--- a/homework/content.js
+++ b/homework/content.js
@@ -4,11 +4,13 @@
   var _wrapper = document.querySelector('#note-content-wrapper');
 
   function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+    var con = new content();
+    window.addEventListener('note-open', 
+	  (function(event) {
+        var note = event.detail;
+        con.resetWrapper();
+        con.drawNote(note);
+      }).bind(this));
   }
 
   function resetWrapper() {
@@ -21,12 +23,12 @@
     h.textContent = title;
     var passages = note.passages;
     var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
+    passages.forEach((function(passage) {
       var p = document.createElement('p');
       p.classList.add('note-passage');
       p.textContent = passage;
       buff.appendChild(p);
-    });
+    }).bind(this));
     _wrapper.appendChild(h);
     _wrapper.appendChild(buff);
   }
--- a/homework/list.js
+++ b/homework/list.js
@@ -6,17 +6,20 @@
   var _wrapper = document.querySelector('#note-list-wrapper');
 
   function start() {
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
+    var li = new list();
+    this.fetchList()
+	.then((function(data) {
+      li.updateList(data);
+      li.drawList();
+      li.preloadFirstNote();
+    }).bind(this));
+    window.addEventListener('click', 
+	  (function(event) {
+        li.onNoteOpen(event);
+      }).bind(this));
   }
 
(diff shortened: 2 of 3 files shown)


homework/content.js
```diff
@@ -4,29 +4,31 @@
   var _wrapper = document.querySelector('#note-content-wrapper');
 
   function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+    var con = new content();
+    window.addEventListener('note-open',
```
snowmantw, line 8: You should bind the listener in argument list with `this`, rather than an anonymous function. See the second example on the MDN page:

homework/content.js
```diff
@@ -34,4 +36,4 @@
   document.addEventListener('DOMContentLoaded', function(event) {
```
snowmantw, line 36: You should remove this start up code and replace it with a centralized one.

homework/list.js
```diff
@@ -6,17 +6,24 @@
...
   function start() {
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
+    var li = new list();
```
snowmantw, line 9: You should **respect your own work**. If you name instances and constructors so arbitrarily like this, no one, include yourself, could maintain it after 3 months, particularly for those projects grow fast. And though it's only a homework, the importance is it would be like a signature shows that you may be a programmer used to write code like this. So, I'm not pissing you off, but I really worry about some tiny issue could be ignored until it affects your opportunities in the future.

snowmantw, line 9: And another issue is you **shouldn't create a instance in a file for definition**. You should leave the work in a file to bootstrap the whole program. One important reason to do that is because the bootstrapping while loading is a proven failed pattern for testing and debugging.

homework/list.js
```diff
@@ -6,17 +6,24 @@
...
+    this.fetchList()
+	.then((function(data) {
```
snowmantw, line 11: You should practice to compose a Promise chain by multiple `then` and `catch`. And you should note that `catch` should **log AND re-throw** the error, or the following chains may hide the message, and then continuously perform some actions that should be ended after one exception occurs.

homework/list.js
```diff
@@ -50,30 +57,35 @@
...
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
+  list.prototype.fetchList = function(afterfetch) {
```
snowmantw, line 65: Since you have successfully "Promisified" the method, you should remove the callback argument.

homework/list.js
```diff
@@ -50,30 +57,35 @@
...
+			}
+		};
+		xhr.send();
+	}).bind(this));
   }
 
   document.addEventListener('DOMContentLoaded', function(event) {
     start();
   });
+  module.exports.add = add;
```
snowmantw, line 89: Did you really test your code? And what made you feel your code is with **good expressiveness and consistent convention** to show your skills? Please, give your work more thoughts. If you have any issue you can't deal with, **ask me**. The only thing worse than making a mistake is that you keep in silent and do not ask for help.

homework/list.js
```diff
@@ -33,16 +40,16 @@
...
 
-  function drawList() {
-    var list = _listNoteContent;
+  list.prototype.drawList = function() {
+    var list2 = _listNoteContent;
     var ul = document.createElement('ul');
     ul.id = 'note-title-list';
     var buff = document.createDocumentFragment();
-    list.forEach(function(note, i) {
+    list2.forEach((function(note, i) {
```
snowmantw, line 52: See? A name collision that should never happen if you give your instance a better name.

homework/content.js
```diff
@@ -4,29 +4,31 @@
   var _wrapper = document.querySelector('#note-content-wrapper');
 
   function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+    var con = new content();
```
snowmantw, line 7: Your naming convention is really not good. See my comments below.

homework/test/test-list.js
```diff
@@ -1,8 +1,10 @@
 describe('Test > ', function() {
   beforeEach(function() {
+  
   });
 
   it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+    assert.equal('2', list.add(1, 1));
```
snowmantw, line 7: I know this looks matching the item in my criteria, but your module and exporting way really needs to be fixed. And, please note you are **comparing a string to a number**, which is dangerous and buggy.

PR #21, opened 2015-10-23
Web_apps HW2 102502511

mike19941215:


homework/content.js
```diff
@@ -27,11 +23,19 @@
...
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
+    this._wrapper.appendChild(h);
+    this._wrapper.appendChild(buff);
+  },
+  
+  start() {
+	  this._wrapper = document.querySelector('#note-content-wrapper');
+      window.addEventListener('note-open', (function(event){
```
snowmantw, line 32: You should set the `this` as the `listener` in the argument list. See the 2nd example of manual here:

homework/list.js
```diff
@@ -1,44 +1,36 @@
...
         { detail: content }));
     };
-  }
+  },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+  preloadFirstNote() {
+    if (this._listNoteContent.length !== 0) {
```
snowmantw, line 21: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,44 +1,36 @@
...
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
-
-  function onNoteOpen(event) {
+TodoListManager.prototype = {
+	
+  onNoteOpen(event) {
```
snowmantw, line 11: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onNoteClick'.

homework/list.js
```diff
@@ -52,28 +44,43 @@
...
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+	}).bind(this));
+  },
+  
+    start() {
+	this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList()
```
snowmantw, line 73: You should practice to compose a promise chain with multiple `then` and `catch`. And you should **log AND re-throw** the error in the `catch`, to prevent the following chains make the error silent and not being interrupted.

homework/test/test-list.js
```diff
@@ -1,8 +1,12 @@
-describe('Test > ', function() {
-  beforeEach(function() {
-  });
+var assert = require('assert');
+var Test = function(){};
+Test.prototype.sub = function (num1,num2){
```
snowmantw, line 3: You should define this function as a method of one manager.

homework/list.js
```diff
@@ -52,28 +44,43 @@
       buff.appendChild(li);
     });
     ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+    this._wrapper.appendChild(ul);
+  },
 
-  function fetchList(afterFetch) {
+  fetchList(afterFetch) {
```
snowmantw, line 50: You have successfully "Promisified" the method, so you should remove the callback argument as well.

PR #22, opened 2015-10-23
Homework refactor

jason1122g:


homework/content.js
```diff
@@ -1,37 +1,78 @@
...
-    });
-  }
+(function(exports) {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
+  /**
+  * Creates an instance of ContentManager.
+  *
+  * @constructor
```
snowmantw, line 8: JSDoc, good work.

homework/content.js
```diff
@@ -1,37 +1,78 @@
...
-    _wrapper.appendChild(buff);
+  ContentManager.prototype = {
+
+    /**
+    * Bind 'note-open' event to initialize the ContentManager.
+    *
+    * @this {ContentManager}
+    */
+    start() {
+      window.addEventListener('note-open', (function(event) {
```
snowmantw, line 23: You need to bind the listener with the `this`, rather than an anonymous function like this.

homework/list.js
```diff
@@ -1,79 +1,126 @@
...
+    */
+    start() {
+      this.fetchList()
+      .then((function(data) {
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this) )
+      .catch((function(){
+        //exception handle code here
```
snowmantw, line 32: **Serious issue**: you can't just capture it and leave nothing. In fact, you need to **log AND re-throw** the error. To log it is to prevent the following chain make the error log silent, while to re-throw it is to make sure the same following chain will be interrupted if there is an error.

homework/list.js
```diff
@@ -1,79 +1,126 @@
...
-    _wrapper.appendChild(ul);
-  }
+    /**
+    * Dispatch a 'note-open' event if event target has 'note-title' in its classList.
+    *
+    * @param {CustomEvent} event Its target may have 'note-title' in classList.
+    * @this {ListManager}
+    */
+    onNoteOpen(event) {
+      if (event.target.classList.contains('note-title')) {
```
snowmantw, line 47: In fact it's a convention to prefix a handler with `on` to the corresponding event. So it should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -1,79 +1,126 @@
...
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
+    /**
+    * Dispatch a 'note-open' event with first note if note list is not empty
+    *
+    * @this {ListManager}
+    */
+    preloadFirstNote() {
+      if (this._listNoteContent.length !== 0) {
```
snowmantw, line 61: If a function body is wrapped with one conditional statement with only one branch, the `if...else` should be put in the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,126 @@
...
+    },
+
+    /**
+    * Fetch remote data of note list.
+    *
+    * @this {ListManager}
+    * @return {Promise} The ajax handler
+    */
+    fetchList() {
+      return new Promise(function( resolve, reject ) {
```
snowmantw, line 108: Good. You have successfully "Promisified" that.

homework/test/test-list.js
```diff
@@ -1,8 +1,16 @@
...
+    contentManager = new ContentManager();
   });
 
-  it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+  it('can calculate the max number', function() {
+    var big   = 10;
+    var small = 5;
+    var maxNumber = contentManager.calculateMaxNumber( big, small );
+    assert.equal( maxNumber, big );
```
snowmantw, line 14: Good. You have done this part.

PR #23, opened 2015-10-23
Homework

StPig:
review=?@snowmantw
review=?@chungya
review=?@FlowerHop

### DIFF
--- a/homework/list.js
+++ b/homework/list.js
@@ -9,7 +9,7 @@
   list.prototype = {
     start() {
       this._wrapper = document.querySelector('#note-list-wrapper');
-      this.fetchList((function(data) {
+      this.fetchList().then((function(data) {
         this.updateList(data);
         this.drawList();
         this.preloadFirstNote();
@@ -59,7 +59,7 @@
     fetchList(afterFetch) {
       return new Promise((function(resolve, reject) {
         var xhr = new XMLHttpRequest();
-        xhr.open('GET', '/demo-list-notes.json', true);
+        xhr.open('GET', 'demo-list-notes.json', true);
         xhr.responseType = 'json';
         xhr.onreadystatechange = function(e) {
           // Watch out: we have a mysterious unknown 'this'.
--- a/homework/list.js
+++ b/homework/list.js
@@ -1,79 +1,81 @@
 'use strict';
 
-(function() {
+(function(exports) {
+  var list = function() {
+    var _listNoteContent = [];
+    var _wrapper = null;
+  };
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  list.prototype = {
+    start() {
+      this._wrapper = document.querySelector('#note-list-wrapper');
+      this.fetchList((function(data) {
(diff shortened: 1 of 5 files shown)


homework/test/test-list.js
```diff
@@ -1,8 +1,13 @@
...
+  var List = [];
   beforeEach(function() {
+    subject = new list();
+    List[1] = "A";
   });
 
-  it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+  it('Test changelist', function() {
+    subject.updateList(List);
```
snowmantw, line 10: It's good to see you test a real method rather than a mocked one.

homework/list.js
```diff
@@ -1,79 +1,81 @@
...
+        buff.appendChild(li);
+      });
+      ul.appendChild(buff);
+      this._wrapper.appendChild(ul);
+    },
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+    fetchList(afterFetch) {
```
snowmantw, line 59: You have successfully "Promisified" the method, so you don't need to keep the callback here. Please remove it and use only Promise interface only.

homework/list.js
```diff
@@ -1,79 +1,81 @@
...
+    var _listNoteContent = [];
+    var _wrapper = null;
+  };
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  list.prototype = {
+    start() {
+      this._wrapper = document.querySelector('#note-list-wrapper');
+      this.fetchList().then((function(data) {
```
snowmantw, line 12: Good, although I would recommend it to use Arrow function instead of `bind` API.

homework/list.js
```diff
@@ -1,79 +1,81 @@
...
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this));
+      window.addEventListener('click', (function(event) {
```
snowmantw, line 17: You need to bind the listener with `this` rather than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,81 @@
...
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+    onNoteOpen(event) {
```
snowmantw, line 22: It's a convention to prefix a handler with `on` before the corresponding event name. So, in fact, this method should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -1,79 +1,81 @@
...
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
-
-  function updateList(list) {
-    _listNoteContent = list;
-  }
+    preloadFirstNote() {
+      if (this._listNoteContent.length !== 0) {
```
snowmantw, line 31: If a function body is wrapped by a conditional statement only with one branch, the `if...else` should be put in the caller, rather than the callee.

PR #25, opened 2015-10-23
webappHW3_update

vickychen95:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,21 +1,17 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
+	var content = function(){
+		var _wrapper=null;
+	}
+	
+  content.prototype={
+	 
+  resetWrapper() {
+    this._wrapper.innerHTML = '';
+  },
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
-
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
-
-  function drawNote(note) {
+  drawNote(note) {
     var title = note.title;
     var h = document.createElement('h2');
     h.textContent = title;
@@ -27,11 +23,20 @@
       p.textContent = passage;
       buff.appendChild(p);
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+    this._wrapper.appendChild(h);
+    this._wrapper.appendChild(buff);
+  },
+  
+  start() {
+		this._wrapper = document.querySelector('#note-content-wrapper');
+		window.addEventListener('note-open', function(event) {
+		var note = event.detail;
+		this.resetWrapper();
+		this.drawNote(note);
+		}).bind(this);
+	}
+  };
+  
+  exports.content=content;
+  
+})(window);
--- a/homework/list.js
+++ b/homework/list.js
@@ -1,44 +1,37 @@
 'use strict';
 
-(function() {
+(function(exports) {
 
+ var list=function(){
   var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  var _wrapper = null;
+ }
+ 
+ list.prototype = {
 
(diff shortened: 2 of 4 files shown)


homework/content.js
```diff
@@ -27,11 +23,20 @@
...
-    start();
-  });
-})();
+    this._wrapper.appendChild(h);
+    this._wrapper.appendChild(buff);
+  },
+  
+  start() {
+		this._wrapper = document.querySelector('#note-content-wrapper');
+		window.addEventListener('note-open', function(event) {
```
snowmantw, line 32: Bind the handler with `this` rather than an anonymous function like this.

homework/list.js
```diff
@@ -1,44 +1,37 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
-
-  function onNoteOpen(event) {
+  onNoteOpen(event) {
     if (event.target.classList.contains('note-title')) {
```
snowmantw, line 13: If a function body is wrapped by a conditional statement only with one branch, the `if...else` should be put in the caller, rather than the callee.

snowmantw, line 13: And in fact it's a convention to prefix `on` to a handler with the corresponding event name. So this method should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -47,33 +40,39 @@
...
     var xhr = new XMLHttpRequest();
     xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
     xhr.responseType = 'json';
     xhr.onreadystatechange = function(e) {
-      // Watch out: we have a mysterious unknown 'this'.
       if (this.readyState === 4 && this.status === 200) {
         var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
+        resolve(listData);
```
snowmantw, line 57: Good. You have successfully "Promisifed" this method with only one nit: you don't need the `afterFetch` callback anymore, so please remove it.

homework/list.js
```diff
@@ -47,33 +40,39 @@
...
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+  }))},
 
-})();
+  start() {
+	this._wrapper = document.querySelector('#note-list-wrapper')
+    this.fetchList(function(data) {
```
snowmantw, line 67: **WRONG**: since you have "Promisifed" the method, you don't need to use the callback. And from this code I bet you didn't test your code by executing it, since it apparently won't work.

Remember: you either use the callback to perform the following steps (the way I require you to refactor), or change it to Promise style and chain the actions with `then` and `catch`. To leave code like this only shows you don't really understand what I've taught you, and you should ask more questions before you submit this. **So I strongly suggest you correct you homework as soon as possible, and ask me for help if you still don't really understand how to use Promise. Otherwise, you may get into trouble**.

homework/main.js
```diff
@@ -0,0 +1,8 @@
+'use strict';  
+   
+ document.addEventListener('DOMContentLoaded', function(event) {  
+     var nContent = new content();
```
snowmantw, line 4: **You need to learn to respect your work, even though it's only a homework**.

To name things like this, and without any comment about the reason why you choose such special convention, is just show you are inexperienced and arbitrary. Although it is true that you are still a student, but please consider that most of your classmates have followed the good convention to complete their homework. **So please rename your constructors and instances with some meaningful and clear names**. If you don't have any idea, check my slides.

homework/list.js
```diff
@@ -1,44 +1,37 @@
...
-(function() {
+(function(exports) {
 
+ var list=function(){
   var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  var _wrapper = null;
+ }
+ 
+ list.prototype = {
```
snowmantw, line 10: **Serious issue**: naming. See my last comment in the `main.js`.

snowmantw:
**Where is your unit test ?**

If there is anything wrong made you miss that part, please correct it **as soon as possible**. And please note that you didn't meet all the criteria I listed for the homework. So please submit another version and set the review again. Otherwise, you'll be failed in this part of the course.

snowmantw:
And if this part is not your whole homework (since you named it as "-update"), please explain it here. You did it confusedly because you only noted the information with a meaningless title and commit message.

PR #28, opened 2015-10-23
102502010

whatword:


homework/content.js
```diff
@@ -27,11 +24,27 @@
...
+        var note = event.detail;
+        this.resetWrapper();
+        this.drawNote(note);
+        break;
+    }
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+    window.addEventListener('note-open', this);
```
snowmantw, line 43: Good. It's always better to bind the `this` as event listener rather than an anonymous function.

homework/list.js
```diff
@@ -1,44 +1,35 @@
...
 
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+  onNoteOpen(event) {
```
snowmantw, line 12: It's a convention to prefix `on` to an event handler with the corresponding event. So this method should be renamed as `onNoteClicked` in fact.

homework/list.js
```diff
@@ -1,44 +1,35 @@
...
+    var content = this._listNoteContent[id];
+    window.dispatchEvent(new CustomEvent('note-open',
+      { detail: content }));
+  },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+  preloadFirstNote() {
+    if (this._listNoteContent.length !== 0) {
```
snowmantw, line 20: If a function body is wrapped by a conditional statement only with one branch, it's better to put the `if...else` in the caller, rather than the callee.

homework/list.js
```diff
@@ -52,28 +43,48 @@
...
-    xhr.responseType = 'json';
-    xhr.onreadystatechange = function(e) {
-      // Watch out: we have a mysterious unknown 'this'.
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
+  fetchList() {
+    return new Promise((function(resolve,reject) {
```
snowmantw, line 50: Good. You have successfully "Promisified" it.

homework/list.js
```diff
@@ -52,28 +43,48 @@
...
+           this.onNoteOpen(event);
+        }
+      break;
+    }
+  },
+
+  start() {
+    window.addEventListener('click',this);
+    this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList().then((function(data) {
```
snowmantw, line 81: In fact, I would recommend to use Arrow function instead of the `bind` API. It's more concise.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
+    beforeEach(function() {
+    	subject = new NoteListManager();
+    });
 
-  it('will test some pure functions', function() {
+  it('It will test drawList functions', function() {
     // Write any pure function assertion here.
+    subject._wrapper = document.createElement('div');
+    subject.drawList([]);
+    assert.isNull(subject._wrapper.querySelector('li'));
```
snowmantw, line 12: Good. You tested a real method instead of a mocked one.

homework/list.js
snowmantw, line 81: Another point is in this way you didn't practice how to chain a Promise. So although it's unnecessary, please try to make a chain with several `then` and a `catch`.

PR #30, opened 2015-10-24
Homework

CaeserNieh:
101502540_pull request

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -4,11 +4,11 @@
   var _wrapper = document.querySelector('#note-content-wrapper');
 
   function start() {
-    window.addEventListener('note-open', function(event) {
+    window.addEventListener('note-open', (function(event) {
       var note = event.detail;
       resetWrapper();
       drawNote(note);
-    }).bind(this);
+    }).bind(this));
   }
 
   function resetWrapper() {
@@ -31,7 +31,7 @@
     _wrapper.appendChild(buff);
   }
 
-  document.addEventListener('DOMContentLoaded', function(event) {
+  document.addEventListener('DOMContentLoaded', (function(event) {
     start();
-  }).bind(this);
+  }).bind(this));
 })();
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,43 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function (exports){
 
-  function start() {
-    window.addEventListener('note-open', (function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    }).bind(this));
+  var ContentManager = function(){
+    this._wrapper = document.querySelector('#note-content-wrapper');
   }
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
+  ContentManager.prototype = {
+    start() {
+        window.addEventListener('note-open', (function(event) {
+              var note = event.detail;
+              this.resetWrapper();
+              this.drawNote(note);
+          }).bind(this));
+    }, 
+    // original content.js start function
+    // call original contest.js resetWrapper function use "this"
+    // following are the same
+    resetWrapper(){
+      _wrapper.innerHTML = '';
+    },
+
+    drawNote(note){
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,43 @@
...
+  var ContentManager = function(){
+    this._wrapper = document.querySelector('#note-content-wrapper');
   }
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+  ContentManager.prototype = {
+    start() {
+        window.addEventListener('note-open', (function(event) {
```
snowmantw, line 11: Instead of binding a function as the listener, you should set the argument as `this` and deal with the event in the `handleEvent` method.

homework/list.js
```diff
@@ -1,43 +1,53 @@
...
+  }
+
+  ListManager.prototype = {
+    start() {
+      this.fetchList().then((function(data) {
+      this.updateList(data);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this))
+    .catch((function () {
```
snowmantw, line 21: **Serious issue**: you need to catch the error in order to **log and re-throw** it. The reason is logging can prevent the following chains ignore it and make troubles when you're debugging, while re-throw it to make sure the next chain will be interrupted, rather than to perform the actions which may lead to unexpected behaviors.

homework/list.js
```diff
@@ -1,43 +1,53 @@
...
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
+  var ListManager = function(){
+    this._listNoteContent = [];
+    this._wrapper = document.querySelector("#note-content-wrapper");
+  }
+
+  ListManager.prototype = {
+    start() {
+      this.fetchList().then((function(data) {
```
snowmantw, line 16: It's recommended to use Arrow function instead of calling `bind` API, since it makes code more concise. See:

homework/list.js
```diff
@@ -1,43 +1,53 @@
...
-  }
+      };
+    }
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
+    preloadFirstNote() {
+      if (_listNoteContent.length !== 0) {
```
snowmantw, line 39: If a function body is wrapped by a conditional statement with only one branch, it's better to move the `if...else` into the caller, rather than the callee.

homework/list.js
```diff
@@ -52,28 +62,26 @@
...
-    _wrapper.appendChild(ul);
+    this._wrapper.appendChild(ul);
   }
 
-  function fetchList(afterFetch) {
-    var xhr = new XMLHttpRequest();
-    xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
-    xhr.responseType = 'json';
-    xhr.onreadystatechange = function(e) {
+  fetchList(afterFetch) {
```
snowmantw, line 68: nit: You have successfully "Promisified" it, so you don't need to keep this callback-style argument.

homework/list.js
```diff
@@ -1,43 +1,53 @@
...
+      this.fetchList().then((function(data) {
+      this.updateList(data);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this))
+    .catch((function () {
+    }).bind(this));
+    
+    window.addEventListener('click', (function(event) {
+      this.onNoteOpen(event).bind(this));
```
snowmantw, line 25: It's a convention to prefix `on` to the method corresponds to the event name. So in fact it should be renamed as `onNoteClicked`.

homework/test/test-list.js
```diff
@@ -2,7 +2,11 @@ describe('Test > ', function() {
   beforeEach(function() {
   });
 
-  it('will test some pure functions', function() {
+  it('my pure testfunctions', function() {
     // Write any pure function assertion here.
+    var a = 1;
+    var b = 2;
+    var sum = a+b;
+    assert.equal(3,sum);
```
snowmantw, line 10: **WRONG**: you need to **define a pure function (method) in one manager, and then call it here.**

PR #32, opened 2015-10-28
101502014 homework

timmycheng1221:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop
Tier 1 finished

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,43 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
-    window.addEventListener('note-open', (function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    }).bind(this));
+  var NoteContentManager = function() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
   }
+  
+  NoteContentManager.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+    start() {
+      window.addEventListener('note-open', (function(event) {
+        var note = event.detail;
+        this.resetWrapper();
+        this.drawNote(note);
+      }).bind(this));
+    },
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+    resetWrapper() {
+      this._wrapper.innerHTML = '';
+    },
 
-  document.addEventListener('DOMContentLoaded', (function(event) {
-    start();
-  }).bind(this));
-})();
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,43 @@
...
+    this._wrapper = document.querySelector('#note-content-wrapper');
   }
+  
+  NoteContentManager.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+    start() {
+      window.addEventListener('note-open', (function(event) {
```
snowmantw, line 12: Instead of binding an anonymous function, you should set the `listener` argument as `this`, and deal with the event in the `handleEvent` method.

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
-        { detail: content }));
-    };
-  }
+    start() {
+      this.fetchList()
+      .then((function(data) {
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this)).catch((function(){
```
snowmantw, line 18: **Serious issue**: you need to catch it, which is correct. But the point is you need to **log AND re-throw** the error in the catch body. To log it is to prevent the following chain will intercept the error and show nothing in your console, which could devastate your debugging strategy wholly. To re-throw it is to make sure the next chain will be interrupted to prevent unexpected behaviors.

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+    onNoteOpen(event) {
+      if (event.target.classList.contains('note-title')) {
```
snowmantw, line 26: If a function body is wrapped by a conditional statement with only one branch, the `if...else` should be moved into the caller, rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
+    },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+    onNoteOpen(event) {
```
snowmantw, line 25: It's a convention to prefix `on` to the method handle the corresponding event. So it should be renamed as `onNoteClicked`.

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
+        // Note: buff is captured, so we now have a
+        // little closure naturally.
+        buff.appendChild(li);
+      });
+      ul.appendChild(buff);
+      this._wrapper.appendChild(ul);
+    },
+
+    fetchList() {
+      return new Promise((function(resolve , reject) {
```

homework/test/test-list.js
```diff
@@ -1,8 +1,22 @@
...
+          "passages": [
+          "An Indian woman who was stranded in Pakistan for a decade has returned home, but says the family she had identified in photos is not hers.",
+          "India's Foreign Minister Sushma Swaraj said the woman, named Geeta, is refusing to recognise her family.",
+          "Geeta arrived in Delhi on Monday morning, days after she identified her family in photos sent from India."
+        ]}
+	];
+
+	test.updateList(dummuyList);
+
+	assert.equal(test._listNoteContent, dummuyList);
```
snowmantw, line 20: It's good to see you test a real method rather than just a pure function.

PR #33, opened 2015-10-28
Homework 102502041

YenchungChen:


homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-        { detail: content }));
-    };
-  }
+  ListManager.prototype = {
+	start(){
+		this.fetchList().then((function(data){
+			this.updateList(data);
+			this.drawList();
+			this.preloadFirstNote();
+		}).bind(this)).catch((function(){
```
snowmantw, line 16: **Serious issue**: if you catch it without **logging AND re-throwing** it, the following chains (if any) will not get interrupted and in the console there will be nothing about the error. It could be a disaster in real programs. So please correct it.

snowmantw, line 16: And the reason we need both logging and re-throwing is because you can prevent the error is covered in the following chains by logging it, while to re-throw it is to make sure the following chain will get interrupted, or some undefined behaviors may cause troubles.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+		}).bind(this));
+		window.addEventListener('click',(function(event){
+			this.onNoteOpen(event);
```
snowmantw, line 20: It's a convention to prefix `on` with a method handling the corresponding event. So it should be renamed as `onClick` or `onNoteClicked` actually. This is a hidden issue about convention and expressiveness.

homework/test/test-list.js
```diff
@@ -1,8 +1,24 @@
...
+		var manager = subject;
+		var dummuyList = [
+			{ "title": "US signals shift in Syria-Iraq campaign against Islamic State",
+			  "passages": [
+			  "The US has indicated a shift in its campaign against Islamic State (IS) militants in Iraq and Syria, including the use of direct ground raids.",
+			  "Defence Secretary Ash Carter said there would also be more air strikes against high-value targets."
+			]}
+		];
+
+		manager.updateList(dummuyList);
```
snowmantw, line 19: It's good to see you test a real instance and method, rather than a mock one.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+		// little closure naturally
+		buff.appendChild(li);
+	});
+	ul.appendChild(buff);
+	this._wrapper.appendChild(ul);
+    },
 
-})();
+    fetchList(){
+	return new Promise(function(resolve, reject){
```
snowmantw, line 64: Good. You have successfully "Promisified" it.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-    xhr.onreadystatechange = function(e) {
-      // Watch out: we have a mysterious unknown 'this'.
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
+  preloadFirstNote(){
+	if(this._listNoteContent.length !== 0){
```
snowmantw, line 34: nit: If a function is only with one branch of conditional statement, the condition should be moved into the caller, rather the callee.

PR #34, opened 2015-10-28
102502505

amm040341:


homework/list.js
```diff
@@ -52,28 +45,48 @@
...
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
-  }
+  fetchList(url) {
+    return new Promise(function(reslove, reject) {
```
snowmantw, line 52: Good. You have successfully "Promisified" it.

homework/test/test-list.js
```diff
@@ -1,8 +1,13 @@
-describe('Test > ', function() {
-  beforeEach(function() {
-  });
+describe('Test Math.sin', function() {
+	var subject;
+	
+	beforeEach(function() {
+		subject = new TestManager;
```
snowmantw, line 5: **WRONG**: why did you think an inexistent instance could be test like this? You should discover the issue if you ever tried to run it. See my file in this repo about how to run the test.

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
+    switch(event.type) {
+      case 'click':
+        this.onNoteOpen(event);
+        break;
+    }
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList('./demo-list-notes.json').then((function(response) {
```
snowmantw, line 80: I recommend to use Arrow function instead of call `bind`:

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
+    });
+  },
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+  handleEvent(event) {
+    switch(event.type) {
+      case 'click':
+        this.onNoteOpen(event);
```
snowmantw, line 73: It's a convention to prefix the `on` to a handler of corresponding event. So it in fact should be renamed as `onNoteClick` or `onClick`. It's a hidden issue about expressiveness.

homework/list.js
```diff
@@ -52,28 +45,48 @@
...
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList('./demo-list-notes.json').then((function(response) {
+      this.updateList(response);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this)).catch(function(error){
+      console.error(error);
```
snowmantw, line 85: It's a serious issue that you catch it but don't **re-throw** it. To log it here is to prevent the following chains make it silent, while to re-throw it is to make sure the following chains outside the method will get interrupted.

homework/content.js
```diff
@@ -27,11 +24,25 @@
...
+          var note = event.detail;
+          this.resetWrapper();
+          this.drawNote(note);
+          break;
+    }
+  },
+
+  start() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+    window.addEventListener('note-open', this);
```
snowmantw, line 43: Good. You noticed it's necessary to bind the `this` as the listener.

PR #35, opened 2015-10-28
102502554

Peter654q:


homework/content.js
```diff
@@ -1,21 +1,26 @@
...
+  var ContentManager = function() {
+  this._wrapper = document.querySelector('#note-content-wrapper');
   }
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+  ContentManager.prototype = {
+  start() {
+    window.addEventListener('note-open', (function(event) {
```
snowmantw: Rather than binding on an anonymous function, you need to bind the `this` via set it as the listener of `addEventListener`.

homework/list.js
```diff
@@ -1,44 +1,36 @@
...
     };
-  }
-
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+  },
+ 
+   preloadFirstNote() {
+    if (this_listNoteContent.length !== 0) {
```
snowmantw: If a function body is wrapped by an `if...else` with only one branch, moving the condition into the caller rather than caller is recommended.

homework/list.js
```diff
@@ -1,44 +1,36 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
-
-  function onNoteOpen(event) {
+ListManager.prototype = {  
+  onNoteOpen(event) {
```
snowmantw: It's a convention to prefix method with `on` and name it as the corresponding event. Therefore, this handler should be renamed as `onClick`.

homework/list.js
```diff
@@ -52,28 +44,41 @@
...
+    return new Promise(function(resolve, reject) {
+		var xhr = new XMLHttpRequest();
+        xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
+        xhr.responseType = 'json';
+        xhr.onreadystatechange = (function(e) {
+          // Watch out: we have a mysterious unknown 'this'.
+          if (this.readyState === 4 && this.status === 200) {
+            var listData = this.response;
+            // The flow ends here.
+            afterFetch(listData);
```
snowmantw: **WRONG**: after wrapping a Promise, you should get rid of the old callback style. This shows you failed to understand what the Promise is, and how to use that. **Please check examples as this to correct your patch**:

And you even didn't check if `afterFetch` exists. It should **never happen** if you ever wrote and ran the code carefully.

homework/list.js
```diff
@@ -52,28 +44,41 @@
...
-    start();
   });
+  },
+  
+  start() {
+      this.fetchList().then((function(data){
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this)).catch((function() {
```
snowmantw: **WRONG**: the similar flow on my slides you may directly refered to **is not the whole story**. To catch an error in a Promise without **re-throwing** it, the following chain will not be interrupted if there is any error. And that will cause some unexpected behaviors which now are worse since you also didn't **log** it. The correct way is you firstly log the error to prevent the following chain might from others ignore it incorrectly just as what you're doing, and then you re-throw it to interrupt the whole chain. If you still have any question, **ask me**, just as I always asked you all to do in the class.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
+
   beforeEach(function() {
+    subject = new TestManager;
   });
 
-  it('will test some pure functions', function() {
+  it('will test the sin function', function(done) {
     // Write any pure function assertion here.
+	subject.set(0);
+    assert.Equal(subject.sin(), 0);
```
snowmantw: **WRONG**: why did you think you can test a manager that **apparently doesn't exist in your patch**? And apparently, you're testing a method **belongs to the built-in Math object** which never is one method of the inexistent manager. Again, it would never happen if you test and write your code carefully. Please correct it.

snowmantw:
You **must** correct your homework as soon as possible, or you really need to perform well in the following classes. If you have any question, **ask me**. I don't mind you ask me 100 times about your difficulties unless you're apparently don't listen to what I answer you. The only thing worse than that is you just keep your homework as this.

Peter654q:
Sorry, I'm not really understand how to use IRC so can I get your email to ask you some questions about javascript?
Cause i'm too shy to ask questions in public , so i want to have some privacy, but I'm really want to do good job in this class , and I really need some help.
Please help me for this work.

snowmantw:
Fair enough. I could understand most of college students are just used to asking no questions because of our culture and education, as a result it's difficult to conquer these puzzles after the class. And that's why I left comments one by one: it's not too late to ask for help and it's not embarrassing, since from my experience people like you can even learn more than your silent classmates, which are with similar problems but just give it up.

Therefore, please don't feel bad, and just try to fix what I've commented. Firstly I think you need to run your program to see what's wrong with that, because obvious errors will emerge in your console. If you don't know how to run the program, here is the link:

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/blob/homework/homework/ABOUT%20HOMEWORK.md

If you're using Windows, python has builds for that (2.x is better):

https://www.python.org/downloads/windows/

And I think you can use the same command in the `cmd.exe`. If there are any unexpected error, you can ask me, and we can find out whether there are other options for that.

You need to run the unit test part with Karma as I post as well:

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/blob/homework/homework/ABOUT%20TEST.md

It needs Node.js and NPM:

https://nodejs.org/en/

Again, if you encounter any issues you can just ask them.

Secondly, about how to ask me questions, I feel that you can just comment on PR page like this. Since it is more convenient to see the code with the comments. Unless you have any issues irrelevant to the code and homework (but relate to coding and our course), you can try to send mail to `gweng@mozilla.com`, that's my company mail and I read it frequently.

PR #36, opened 2015-10-28
102502035

lihungte96:
Lesson-03-JavaScriptBestPratice-HW homework

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -8,8 +8,8 @@
         start() {
             window.addEventListener('note-open', function (event) {
                 var note = event.detail;
-                resetWrapper();
-                drawNote(note);
+                this.resetWrapper();
+                this.drawNote(note);
             });
             this._wrapper = document.querySelector('#note-content-wrapper');
         },
--- a/homework/content.js
+++ b/homework/content.js
@@ -32,5 +32,5 @@
             _wrapper.appendChild(buff);
         }
     };
-    exports.TodoListManager = TodoListManager;
+    exports.TodoContentManager = TodoContentManager;
 })(window);
--- a/homework/content.js
+++ b/homework/content.js
@@ -6,12 +6,17 @@
     }
     TodoContentManager.prototype = {
         start() {
-            window.addEventListener('note-open', function (event) {
-                var note = event.detail;
-                this.resetWrapper();
-                this.drawNote(note);
-            });
+            window.addEventListener('note-open', this);
             this._wrapper = document.querySelector('#note-content-wrapper');
+        },
+        handleEvent(event) {
+            switch (event.type) {
+                case 'note-open':
+                    var note = event.detail;
+                    this.resetWrapper();
+                    this.drawNote(note);
+                    break;
+            }
         },
         resetWrapper() {
             this._wrapper.innerHTML = '';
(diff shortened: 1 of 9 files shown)


homework/content.js
```diff
@@ -1,37 +1,71 @@
...
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+(function (exports) {
+    /**
```
snowmantw, line 4: It's good to see you tried to JSDoc the code.

homework/content.js
```diff
@@ -1,37 +1,71 @@
...
+    TodoContentManager.prototype = {
+        /**
+         * Initiate wrapper.
+         * Add a event listener to listen to note-open event
+         *
+         * @this {TodoContentManager}
+         */
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', this);
```
snowmantw, line 22: Good. Bind `this` as the listener is better than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,142 @@
...
+         * Add a eventlistener to listen click event 
+         * Call fetchList to fetech data,
+         * after fetch , draw the list
+         * 
+         * @this {TodoListManager}
+         */
+        start() {
+            this._wrapper = document.querySelector('#note-list-wrapper');
+            window.addEventListener('click', this);
+            this.fetchList().then((function (data) {
```
snowmantw, line 29: I recommend to use Arrow function instead of call `bind` API. It's more expressive, see:

homework/list.js
```diff
@@ -1,79 +1,142 @@
...
+        },
+        /**
+         * Dispatch a custom event when mouse click at the note-title.
+         * Also determine the content to be showed.
+         *
+         * @param {Object} event- mouse click event.
+         * @this {TodoListManager}
+         */
+        onNoteOpen(event) {
+            if (event.target.classList.contains('note-title')) {
```
snowmantw, line 43: If a function body is wrapped by an `if...else` with only one branch, moving the condition into the caller rather than caller is recommended.

homework/list.js
```diff
@@ -1,79 +1,142 @@
...
+         * If there is any problem fetch the file direct from same http server as .html file
+         * (Because I got some problem to get the file from my system)
+         * If success,then continue the following flow 
+         * Else it will reject
+         *
+         * @return {Promise}
+         * @this {TodoListManager}
+         */
+        fetchList() {
+            return new Promise((function (yes, no) {
```
snowmantw, line 118: Well, if you work along this may not be a problem. But to keep the documented names of arguments make your code more readable for others. Anyway, you have successfully "Promisified" the method, good work.

homework/test/test-list.js
```diff
@@ -2,7 +2,8 @@ describe('Test > ', function() {
   beforeEach(function() {
   });
 
-  it('will test some pure functions', function() {
+  it('will test some pure functions', function () {
+      assert.deepEqual(test.test(1,1), 2);
```
snowmantw, line 6: Well, you may need to check if you did successfully include the `test.js` and test it here.

PR #37, opened 2015-10-29
101502522 HW3

weiyuan1993:
101502522

### DIFF
--- a/homework/list.js
+++ b/homework/list.js
@@ -1,79 +1,90 @@
 'use strict';
 
 (function() {
+  
+  var ListManager = function(){
+    this._listNoteContent = [];
+    this._wrapper = null;
+  };
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+  ListManager.prototype=
+  {
+     drawList() {
+      var list = this._listNoteContent;
+      var ul = document.createElement('ul');
+      ul.id = 'note-title-list';
+      var buff = document.createDocumentFragment();
+      list.forEach(function(note, i) {
+        var li = document.createElement('li');
+        li.dataset.noteId = i;
+        li.classList.add('note-title');
+        li.textContent = note.title;
+        // Note: buff is captured, so we now have a
+        // little closure naturally.
+        buff.appendChild(li);
+      });
+      ul.appendChild(buff);
+      this._wrapper.appendChild(ul);
+    },
 
-  function start() {
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
 
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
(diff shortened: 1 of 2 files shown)


weiyuan1993:
101502522

homework/list.js
```diff
@@ -1,79 +1,90 @@
...
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+    onNoteOpen(event) {
+      if (event.target.classList.contains('note-title')) {
```
snowmantw, line 32: If a function is wrapped in a condition statement with only one branch, it's better to move the `if...else` into the caller, rather than callee.

homework/list.js
```diff
@@ -1,79 +1,90 @@
...
+      return new Promise((function(resolve,reject){
+      var xhr = new XMLHttpRequest();
+      xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
+      xhr.responseType = 'json';
+      xhr.onreadystatechange = function(e) {
+        // Watch out: we have a mysterious unknown 'this'.
+        if (this.readyState === 4 && this.status === 200) {
+          var listData = this.response;
+          // The flow ends here.
+          afterFetch(listData);
```
snowmantw, line 62: **WRONG**: You either keep it as a callback-style function (so you **won't** get the point) OR re-write it as a Promise-based function (that's what I need you to complete). But to wrap it inside a Promise **without resolving or rejecting the result** is showing you didn't understand Promise after all. Please, check the MDN link here:

Or if you like, you can look up any other examples on the Internet.

homework/list.js
```diff
@@ -1,79 +1,90 @@
...
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
-  }
+    start() {
+      this.fetchList((function(data) {
```
snowmantw, line 72: It's a crucial mistake here, and again shows you did completely not know what I've taught. I believe I already asked all of you lots of times whether you had questions or not, and this error just shows you might need to ask more, if you still don't know why this is a mistake.

homework/list.js
```diff
@@ -1,79 +1,90 @@
...
+          this.drawList();
+          this.preloadFirstNote();
+        }).bind(this));
+      window.addEventListener('click', (function(event) {
+        this.onNoteOpen(event);
+      }.bind(this)));
+    }
+  };
 
   document.addEventListener('DOMContentLoaded', function(event) {
```
snowmantw, line 83: **Don't keep this trick used in the (bad) closure example**. You need to centralize them into one start-up method.

snowmantw:
Don't you think you also need to refactor other files in the `homework` directory?

I strongly suggest you doing these as soon as possible to avoid getting into trouble:
1. Review and try what I've taught in the class
2. If you have any question, ask me
3. Take a look at what my homework criteria is
4. Correct your homework according to that one-by-one
5. Update your PR again and info me by set review again

PR #38, opened 2015-10-29
Homework 101502017

nickyeh97:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -4,7 +4,7 @@
   
   var ToDoContentManager = function () {
       this._wrapper = null;  
-  }
+  };
   ToDoContentManager.prototype = {
     start(){
         /*
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,45 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
-
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+(function(exports) {
+  
+  var ToDoContentManager = function () {
+      this._wrapper = null;  
   }
+  ToDoContentManager.prototype = {
+    start(){
+        /*
+         *Initiate wrapper.
+         *To do bind(this)
+         */
+        this._wrapper = document.querySelector('note-content-wrapper');
+        window.addEventListener('note-open',(function(event){
+          var note = event.detail;
+          this.resetWrapper();
+          this.drawNote(note);
+        }).bind(this));
+    },
+      
+    resetWrapper() {
+      this._wrapper.innerHTML = '';
+    },
+    
+    drawNote(note) {
+      var title = note.title;
+      var h = document.createElement('h2');
+      h.textContent = title;
+      var passages = note.passages;
+      var buff = document.createDocumentFragment();
+      passages.forEach(function(passage) {
+        var p = document.createElement('p');
+        p.classList.add('note-passage');
+        p.textContent = passage;
+        buff.appendChild(p);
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,45 @@
...
+      this._wrapper = null;  
+  };
+  ToDoContentManager.prototype = {
+    start(){
+        /*
+         *Initiate wrapper.
+         *To do bind(this)
+         */
+        this._wrapper = document.querySelector('note-content-wrapper');
+        window.addEventListener('note-open',(function(event){
```
snowmantw, line 15: Rather than binding on an anonymous function, you need to bind the `this` via set it as the listener of `addEventListener`.

homework/list.js
```diff
@@ -1,44 +1,48 @@
...
+  
+  ToDoListManager.prototype = {
+    start(){
+      this.fetchList()
+      .then((function(data) {
+        this.updateList(data);
+        this.drawList();
+        this.preloadFirstNote();
+      }).bind(this));
+      window.addEventListener('click', (function(event) {
```
snowmantw, line 18: It's a convention to prefix `on` to the corresponding function. So it should be renamed as `onClick` actually.

homework/list.js
```diff
@@ -1,44 +1,48 @@
...
+  
+  var ToDoListManager = function(){
+    this._listNoteContent = [];
+    this._wrapper = document.querySelector('#note-list-wrapper');
+  };
+  
+  ToDoListManager.prototype = {
+    start(){
+      this.fetchList()
+      .then((function(data) {
```
snowmantw, line 13: I recommend you to use Arrow function instead of call `bind`, since it's more expressive. See:

homework/list.js
```diff
@@ -52,28 +56,30 @@
...
-      }
-    };
-    xhr.send();
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+    fetchList(afterFetch) {
+      return new Promise((function(resolve , reject){
```
snowmantw, line 63: It's good to see you have successfully "Promisified" the function. However, please remove the callback argument. It's now useless.

homework/main.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+document.addEventListener('DOMContentLoaded',function(event){
+  var  ToDoContentManager = new ToDoContentManager();
+  var  ToDoListManager = new ToDoListManager();
+  ToDoContentManager.start();
```
snowmantw, line 6: Please don't overwrite existing classes with the instances. You should rename them as `toDoContentManager` and `toDoListManager`.

homework/test/test-list.js
```diff
@@ -1,8 +1,12 @@
...
+describe('Test ToDoListManager ', function() {
+  var subject;
   beforeEach(function() {
+      subject = new ToDoListManager();
   });
 
   it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+    var test = subject;
+    var sum = test.puerFunction(1,1);
```
snowmantw, line 9: Where is your `puerFunction`? And a typo?

homework/list.js
```diff
@@ -1,44 +1,48 @@
...
-  }
-
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+      };
+    },
+    
+    preloadFirstNote(){
+      if (this._listNoteContent.length !== 0) {
```
snowmantw, line 33: If a function is wrapped in a condition statement with only one branch, it's better to move the `if...else` into the caller, rather than callee.

PR #39, opened 2015-10-29
102502557

Davisanity:


homework/content.js
```diff
@@ -1,37 +1,39 @@
...
-  }
+(function(exports) {                                 //exports是自己命名的 其實就是window
+  var ListContentManager = function(){
+    this._wrapper = null
+  });
+ 
+  ListNoteContentManager.prototype = {
+      start() {
+              this._wrapper = document.querySelector('#note-content-wrapper');
+              window.addEventListener('note-open', (function(event) {
```
snowmantw, line 11: Instead of binding an anonymous function, you should bind the this as the listener. And Arrow function is better than the binding API (it's more concise), although I didn't require you to use that. See:

homework/index.html
```diff
@@ -3,8 +3,9 @@
 <head>
   <meta charset="UTF-8">
   <title> Homework - Note List </title>
-  <script defer src="list.js"></script>
-  <script defer src="content.js"></script>
+  <script src="list.js"></script>
```
snowmantw, line 6: To set `defer` is to prevent loading JS blocks HTML loading, so you should keep it.

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+  onNoteOpen(event) {
```
snowmantw, line 22: It's a convention to prefix method with `on` for the corresponding function, so in fact it should be renamed as 'onClick'

homework/list.js
```diff
@@ -1,79 +1,87 @@
...
+            buff.appendChild(li);
+        });
+        ul.appendChild(buff);
+        this._wrapper.appendChild(ul);
+    },
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
+   fetchList() {
```

homework/test/test-list.js
```diff
@@ -1,8 +1,13 @@
...
+    var subject;
+    beforeEach(function() {
+        subject = new ListNoteContentManager();
+    });
+it('Test resetWrapper function', function() {
+        // Write any pure function assertion here.
+        subject._wrapper = document.createElement('div');
+        subject._wrapper.innerHTML = 'Test';
+        subject.resetWrapper();
+        assert.equal(subject._wrapper.innerHTML, '', "Reset Failed!");
```
snowmantw, line 11: It's good to see you test a real method rather than a simple one.

PR #40, opened 2015-10-29
Homework 102502043

kaofushwai:


homework/content.js
```diff
@@ -1,37 +1,40 @@
...
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+  ContentManager.prototype = {
+	start(){
+		window.addEventListener('note-open', (function(event){
```
snowmantw, line 11: Instead of binding an anonymous function, you should bind the `this` as the listener. And Arrow function is better than the binding API (it's more concise), although I didn't require you to use that. See:

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+	var ListManager = function();
+	 this._listNoteContent = [];
+	 this._wrapper = document.querySelector('#note-list-wrapper');
   }
 
-  function updateList(list) {
-    _listNoteContent = list;
-  }
+  ListManager.prototype = {
+  start(){
```
snowmantw, line 11: nit: I just wonder why your indentation looks so weird...

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+  this.fetchList()
+  .then((function)(data){
+  this.updateList(data);
+  this.drawList();
+  this.preloadFirstNote();
+  }).bind(this))
+  .catch((function(){
```
snowmantw, line 18: Serious issue here: please remember to log **AND** re-throw Promise error. To log it is because you can know the error early and in your own format (also, to prevent it being captured improperly by the following chain), and to re-throw it is to prevent the other chains concating it keep running after the error.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+	 this._wrapper = document.querySelector('#note-list-wrapper');
   }
 
-  function updateList(list) {
-    _listNoteContent = list;
-  }
+  ListManager.prototype = {
+  start(){
+  this.fetchList()
+  .then((function)(data){
```
snowmantw, line 13: `(function)(data)`

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-      li.classList.add('note-title');
-      li.textContent = note.title;
-      // Note: buff is captured, so we now have a
-      // little closure naturally.
-      buff.appendChild(li);
-    });
-    ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+  onNoteOpen(event){
```
snowmantw, line 27: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+   if(event.target.classList,contains('note-title')){
+   var id = event.target.dataset.noteId;
+   vat content = this._listNoteContent[id];
+   window.dispatchEvent(new CustomEvent('note-open',
+	detail: content}));
+	};
+ },
+  
+  preloadFirstNote(){
+	if(this.listNoteContent.lenght !== 0){
```
snowmantw, line 37: If a function is only with one `if...else` and there is no `else` branch after all, it's better to move the condition out of the callee.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
+  drawList(){
```
snowmantw, line 48: You really need to correct your indentation...

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+  li.classList.add('note-title');
+  li.textContent = note.title;
+  buff.appendChild(li);
   });
+  ul.appendChild(buff);
+  this._wrapper.appendChild(ul);
+  },
+  
+  fetchList(){
+	return new Promise(function(resolve, reject){
```

homework/test/test-list.js
```diff
@@ -1,8 +1,13 @@
...
+describe('Test Math',function(){
   beforeEach(function() {
   });
 
-  it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+  it('Calculate the Maximum num', function() {
+    var small = 1;
+	var big = 10;
+	var maxNumber = Math.max(big,small);
```
snowmantw, line 10: It's better to write a method of manager and then call it here.

PR #41, opened 2015-10-29
102502047

jack99541008:
@snowmantw  @chungya  @FlowerHop 
I do something about these:
1.Bind this in Event handlers
2.Deal with object properties w/ constructor & prototype
3.Deal with asynchronous flows with Promise
4.Write my first test case for a pure function

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,21 +1,24 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
-
-  function start() {
-    window.addEventListener('note-open', function(event) {
+(function(exports) {
+  var ContentManager = function() {
+     this.wrapper = document.querySelector('#note-content-wrapper');
+  }
+  
+ ContentManager.prototype = {
+  start() {
+    window.addEventListener('note-open', (function(event) {
       var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+      this.resetWrapper();
+      this.drawNote(note);
+    }).bind(this));
+  },
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+  resetWrapper() {
+    this.wrapper.innerHTML = '';
+  },
 
-  function drawNote(note) {
+  drawNote(note) {
     var title = note.title;
     var h = document.createElement('h2');
     h.textContent = title;
@@ -27,11 +30,12 @@
       p.textContent = passage;
       buff.appendChild(p);
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
+    this.wrapper.appendChild(h);
+    this.wrapper.appendChild(buff);
   }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
+ };
+ /* document.addEventListener('DOMContentLoaded', function(event) {
     start();
-  });
-})();
+  });*/
+exports.ContentManager = ContentManager;
+})(window);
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,21 +1,24 @@
...
-      drawNote(note);
-    });
+(function(exports) {
+  var ContentManager = function() {
+     this.wrapper = document.querySelector('#note-content-wrapper');
   }
+  
+ ContentManager.prototype = {
+  start() {
+    window.addEventListener('note-open', (function(event) {
```
snowmantw, line 10: It is better to bind `this` as the listener, rather than an anonymous function.

homework/content.js
```diff
@@ -27,11 +30,12 @@
...
     });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
+    this.wrapper.appendChild(h);
+    this.wrapper.appendChild(buff);
   }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
+ };
+ /* document.addEventListener('DOMContentLoaded', function(event) {
```
snowmantw, line 37: You can just remove unnecessary code.

homework/list.js
```diff
@@ -1,44 +1,51 @@
...
-    });
-  }
+ListManager.prototype = {
+
+  start() {
+      this.fetchList().then((function(response) {
+      this.updateList(response);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this)).catch((function() {
```
snowmantw, line 17: Serious issue: don't catch error without logging AND re-throwing it, since this will make the error disappear silently. Consider what will happen if the chain contains an error but your program won't alert anything about that.

homework/list.js
```diff
@@ -1,44 +1,51 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+ListManager.prototype = {
+
+  start() {
+      this.fetchList().then((function(response) {
```
snowmantw, line 13: nit: I recommend to use Arrow function rather than just function binding, although they're both legit.
See:

homework/list.js
```diff
@@ -1,44 +1,51 @@
...
         { detail: content }));
     };
-  }
+  },
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
+  preloadFirstNote() {
+    if (this._listNoteContent.length !== 0) {
```
snowmantw, line 36: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -52,28 +59,31 @@
...
 
-  function fetchList(afterFetch) {
-    var xhr = new XMLHttpRequest();
-    xhr.open('GET', 'http://127.0.0.1:8000/demo-list-notes.json', true);
-    xhr.responseType = 'json';
-    xhr.onreadystatechange = function(e) {
-      // Watch out: we have a mysterious unknown 'this'.
-      if (this.readyState === 4 && this.status === 200) {
+  fetchList() {
+	return new Promise((function(resolve, reject) {
```
snowmantw, line 66: Good. You have successfully "Promisified" this method.

homework/test/test-list.js
```diff
@@ -4,5 +4,6 @@ describe('Test > ', function() {
 
   it('will test some pure functions', function() {
     // Write any pure function assertion here.
+    assert.equal(1+2, 3);
```
snowmantw, line 7: No. You need to create a pure method in one manager, and then test to invoke it here.

PR #42, opened 2015-10-29
102502559

CrashedBboy:
## Homework3

Javascript code refactoring

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,41 +1,39 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function Note() {
+  this._wrapper = document.querySelector('#note-content-wrapper');
+  document.addEventListener('DOMContentLoaded', this.start.bind(this));
+}
 
-  function start() {
-		// 'note-open' will be triggered at list.js
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+Note.prototype.start = function() {
+	// 'note-open' will be triggered at list.js
+	window.addEventListener('note-open', this.updateContent.bind(this));
+}
 
-	/** Clear content container for the new one */
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+Note.prototype.updateContent = function(event) {
+	var note = event.detail;
+	this.resetWrapper();
+	this.drawNote(note);
+}
 
-	/** Append the new one */
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
+Note.prototype.resetWrapper = function() {
+	this._wrapper.innerHTML = '';
+}
 
-	/** which start() will be executed first? */
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+Note.prototype.drawNote = function(note) {
+	var title = note.title;
+	var h = document.createElement('h2');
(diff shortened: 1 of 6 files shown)


homework/content.js
```diff
@@ -1,37 +1,53 @@
...
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+
+  /**
+  * Create a instance of contentManager.
+  *
+  * @constructor
+  */
```
snowmantw, line 9: Good to see you tried to JSDoc it.

homework/content.js
```diff
@@ -1,37 +1,53 @@
...
+	var ContentManager = function() {
+		/** @private */ this._wrapper = document.querySelector('#note-content-wrapper');
+		document.addEventListener('DOMContentLoaded', this.start.bind(this));
+	}
+
+	ContentManager.prototype = {
+
+    /** Initialization */
+		start() {
+			window.addEventListener('note-open', this.updateContent.bind(this));
```
snowmantw, line 19: To set `this` as the listener is better than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
-    });
-  }
+		/** Initialize the listManager. */
+		start() {
+			this.fetchList()
+				.then(this.updateList.bind(this))
+				.then(this.drawList.bind(this))
+				.then(this.preloadFirstNote.bind(this))
+				.catch(function(error) {
+					console.error(error);
```
snowmantw, line 24: It's necessary to re-throw the error after logging it. Since you will silent the error after this step.

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+		/** Initialize the listManager. */
+		start() {
+			this.fetchList()
```
snowmantw, line 19: Good to see you have "Promisified" the chain and use it with the following one.

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+		/**
+    * Get List form localhost
+    *
+    * @return {Promise} A start point of the work flow.
+    */
+		fetchList(afterFetch) {
```
snowmantw, line 34: Since you have already "Promisified" the function, the callback argument here is unnecessary.

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
-      // Note: buff is captured, so we now have a
-      // little closure naturally.
-      buff.appendChild(li);
-    });
-    ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+		/** Show first note before user do anything. */
+		preloadFirstNote() {
+			if (this._listNoteContent.length !== 0) {
```
snowmantw, line 68: To improve expressiveness you should move the condition outside the callee, and put it in the caller. Since there is only one branch of such `if...else`

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
-    };
-    xhr.send();
-  }
+    /**
+    * Trigger 'note-open' event while note opened
+    *
+    * @this {ListManager}
+    * @param {object} event The information such as event source, which is about event.
+    */
+		onNoteOpen(event) {
```
snowmantw, line 81: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'

homework/test/test-list.js
```diff
@@ -3,6 +3,6 @@ describe('Test > ', function() {
   });
 
   it('will test some pure functions', function() {
-    // Write any pure function assertion here.
+		expect(true).to.be.true;
```
snowmantw, line 6: No. You should test a method defined in one manager.

homework/list.js
```diff
@@ -1,79 +1,99 @@
...
+	}
+	exports.ListManager = ListManager;
+})(window);
 
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-
-})();
+function insertItem(element, index) {
```
snowmantw, line 93: Although it's a good try to create a method within one specific scope, it's still weird to see this in a manager styled file.

PR #43, opened 2015-10-29
Homework

w181496:
review?=@snowmantw,@chungya,@FlowerHop
Tier 1 done

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,42 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+  var ContentManager = function() {
+    this._wrapper = null;
+  };
+
+  ContentManager.prototype = {
+    start() {
+      this._wrapper = document.querySelector('#note-content-wrapper');
+      window.addEventListener('note-open', this);
+      /*(function(event) {
+        var note = event.detail;
+        resetWrapper();
+        drawNote(note);
+      }).bind(this));
+      */
+    },
+
+    resetWrapper() {
+      this._wrapper.innerHTML = '';
+    },
+
+    drawNote(note) {
+      var title = note.title;
+      var h = document.createElement('h2');
+      h.textContent = title;
+      var passages = note.passages;
+      var buff = document.createDocumentFragment();
+      passages.forEach(function(passage) {
+        var p = document.createElement('p');
+        p.classList.add('note-passage');
+        p.textContent = passage;
+        buff.appendChild(p);
+      });
+      this._wrapper.appendChild(h);
+      this._wrapper.appendChild(buff);
+    }
   }
-
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
-
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,42 @@
...
+    this._wrapper = null;
+  };
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+  ContentManager.prototype = {
+    start() {
+      this._wrapper = document.querySelector('#note-content-wrapper');
+      window.addEventListener('note-open', this);
```
snowmantw, line 12: Good to see you bound an instance (`this`) instead of binding a function.

homework/content.js
```diff
@@ -1,37 +1,42 @@
...
+      /*(function(event) {
```
snowmantw, line 13: nit: Commented code could cause troubles. Remove them.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+  ListManager.prototype = {
+    start() {
+      this._wrapper = document.querySelector('#note-list-wrapper');
+      this.fetchList().then((function(data) {
```
snowmantw, line 12: I recommend to use Arrow function: `() => { }` instead of binding this manually.
See:

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
 
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+      window.addEventListener('click', (function(event) {
```
snowmantw, line 18: While you use binding this in the previous `addEventListener`, you should keep the convention consistent. And, It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
-      li.textContent = note.title;
-      // Note: buff is captured, so we now have a
-      // little closure naturally.
-      buff.appendChild(li);
-    });
-    ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+    onNoteOpen(event) {
+      if (event.target.classList.contains('note-title')) {
```
snowmantw, line 24: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+        // Note: buff is captured, so we now have a
+        // little closure naturally.
+        buff.appendChild(li);
+      });
+      ul.appendChild(buff);
+      this._wrapper.appendChild(ul);
+    },
+
+    fetchList(afterFetch) {
+      return new Promise((function(resolve, reject) {
```
snowmantw, line 63: Good. You have successfully "Promisified" it.

homework/test/test-list.js
```diff
@@ -1,8 +1,14 @@
...
+    sub = new ContentManager();
   });
 
-  it('will test some pure functions', function() {
+  it('test reset wrapper', function() {
     // Write any pure function assertion here.
+    sub._wrapper = document.createElement('p');
+    sub._wrapper.innerHTML = "G_G";
+    sub.resetWrapper();
+    assert.equal(sub._wrapper.innerHTML , "" ,"Fail！")
```
snowmantw, line 12: It's good to see you tested a real method rather than a simple function.

PR #44, opened 2015-10-29
102502524

esxleader:


homework/content.js
```diff
@@ -1,21 +1,26 @@
...
-      drawNote(note);
-    });
-  }
+ContentManager.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
+  start() {
+	  this._wrapper = document.querySelector('#note-content-wrapper');
+      window.addEventListener('note-open', (function(event){
```
snowmantw, line 12: Instead of using an anonymous function, setting the listener as `this` and handle events with `handleEvent` method is better.

homework/list.js
```diff
@@ -1,44 +1,49 @@
...
-    });
-  }
-
-  function onNoteOpen(event) {
+ListManager.prototype = {
+	
+  start() {
+	this._wrapper = document.querySelector('#note-list-wrapper');
+    this.fetchList()
+	.then((function(data) {
```
snowmantw, line 14: Okay, good. You have successfully "Promisified" it and use it with another chain.

snowmantw, line 14: However, I suggest Arrow function is better than using the `bind` API. See this:

homework/list.js
```diff
@@ -1,44 +1,49 @@
...
+      this.updateList(data);
+      this.drawList();
+      this.preloadFirstNote();
+    }).bind(this));
+    window.addEventListener('click', (function(event) {
```
snowmantw, line 19: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'. And, to set `this` as the listener is better than an anonymous function.

homework/list.js
```diff
@@ -52,10 +57,11 @@
       buff.appendChild(li);
     });
     ul.appendChild(buff);
-    _wrapper.appendChild(ul);
-  }
+    this._wrapper.appendChild(ul);
+  },
 
-  function fetchList(afterFetch) {
+  fetchList(afterFetch) {
```
snowmantw, line 63: Since you have "Promisified" it, to keep this callback argument is unnecessary.

homework/test/test-list.js
```diff
@@ -1,8 +1,13 @@
...
-  beforeEach(function() {
-  });
+var assert = require('assert');
+var Test = function(){};
 
-  it('will test some pure functions', function() {
-    // Write any pure function assertion here.
-  });
-});
+Test.prototype.add = function (n1,n2){
```
snowmantw, line 4: It's better to define it in one of the managers, rather than to write it in the test file.

PR #45, opened 2015-10-30
101502012

Danny05922161:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop

### DIFF
--- a/homework/list.js
+++ b/homework/list.js
@@ -1,11 +1,15 @@
 'use strict';
 
-(function() {
-
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
-
-  function start() {
+(function(exports) {
+  
+  var TodoListManager = function() {
+    this._listNoteContent = [];
+    this._wrapper = document.querySelector('#note-list-wrapper');
+  };
+  
+TodoListManager.prototype = {
+  
+  start() {
     fetchList(function(data) {
       updateList(data);
       drawList();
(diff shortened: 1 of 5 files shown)


DannyJJJJ:
test change

homework/list.js
```diff
@@ -1,44 +1,48 @@
...
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
-
-  function onNoteOpen(event) {
+  onNoteOpen(event) {
```
snowmantw, line 23: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'.

homework/test/test-list.js
```diff
@@ -1,8 +1,17 @@
...
   });
 
   it('will test some pure functions', function() {
     // Write any pure function assertion here.
+    var test = subject;
+    var sum = test.puerFunction(1,4);
+    assert.equal(sum,5,'It is equal to 5!');
   });
+  
+  Manager.prototype.puerFunction(a,b){
```
snowmantw, line 14: 1. There is no `Manager` constructor in this file and the included files
2. Put it in any manager you have defined
3. Are you sure this works well?

homework/list.js
```diff
@@ -52,28 +56,28 @@
...
-      if (this.readyState === 4 && this.status === 200) {
-        var listData = this.response;
-        // The flow ends here.
-        afterFetch(listData);
-      } else if (this.status !== 200 ){
-        // Ignore error in this case.
-      }
-    };
-    xhr.send();
+  fetchList(afterFetch) {
```
snowmantw, line 62: It's **wrong** here: if you are using a Promise, you **definitely** needn't a callback-style method. Instead of, you return a Promise. Please check Promise example like this:

 I suggest that you better to check and correct it, or you're just showing that you haven't understood Promise well.

snowmantw:
After check, your critical error **look the same** as 102502038's:

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/pull/53

Please correct it **by yourself**.

PR #46, opened 2015-10-30
Homework

hsuehkuan-lu:
Finish HW3 Tier 1


hsuehkuan-lu:

hsuehkuan-lu:
100502203

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+  // constructor
+  var self = function () {
```
snowmantw, line 6: It may not need to create a inner function as the constructor, while the `ContentManager` **is** the constructor. And to overwrite the keyword `self` is not so good.

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
+  };
+ 
+  // prototype
+  self.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+    start() {
+      this._wrapper = document.querySelector('#note-content-wrapper');
```
snowmantw, line 14: Instead of using an anonymous function, setting the listener as `this` and handle events with `handleEvent` method is better.

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
+      .then(this.handler.bind(this));
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+      window.addEventListener('click', (function(event) {
```
snowmantw, line 20: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'. And, to set `this` as the listener is better than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
+        li.textContent = note.title;
+        // Note: buff is captured, so we now have a
+        // little closure naturally.
+        buff.appendChild(li);
+      });
+      ul.appendChild(buff);
+      this._wrapper.appendChild(ul);
+    },
+
+    fetchList(afterFetch) {
```
snowmantw, line 70: Good, you have successfully "Promisified" it.

hsuehkuan-lu:
Just realized I used to learn the concept of `Promise` before, totally forgot the knowledge of Javascript!

PR #47, opened 2015-10-30
101502018

real3856:
review=?@snowmantw

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,42 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function Content() {
+	this._wrapper = document.querySelector('#note-content-wrapper');
+}
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+Content.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+	start : function () {
+		window.addEventListener('note-open', (function(event) {
+		var note = event.detail;
+		this.resetWrapper();
+		this.drawNote(note);
+		}).bind(this));
+	},
+	resetWrapper : function () {
+		this._wrapper.innerHTML = '';
+	},
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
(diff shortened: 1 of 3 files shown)


homework/content.js
```diff
@@ -1,37 +1,42 @@
...
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+Content.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+	start : function () {
```
snowmantw, line 9: Instead of using an anonymous function, set the listener as `this` and handle events with `handleEvent` method is better.

homework/content.js
```diff
@@ -1,37 +1,42 @@
...
+		});
+		this._wrapper.appendChild(h);
+		this._wrapper.appendChild(buff);
+	},
+	addition: function (a,b) {
+	    return a + b;
+	}
+};
+var content = new Content();
+document.addEventListener('DOMContentLoaded', function(event) {
```
snowmantw, line 40: It's better to put all startup code like this in one method, rather than two native handlers like this.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
 
-(function() {
+function List() {
+	this._listNoteContent = [];
+	this._wrapper = document.querySelector('#note-list-wrapper');
+}
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+List.prototype = {
```
snowmantw, line 8: You can re-write the chain as

```
this.fetchList().then(this.updateList.bind(this))
                      .then(this.drawList.bind(this))
```

It's more concise.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
-    fetchList(function(data) {
-      updateList(data);
-      drawList();
-      preloadFirstNote();
-    });
-    window.addEventListener('click', function(event) {
-      onNoteOpen(event);
-    });
-  }
+	onNoteOpen : function (event) {
```
snowmantw, line 22: Well it's a convention to prefix with `on` to handle the corresponding event, rather than to fire it. So, you should rename this function as `onClick` and to dispatch the `note-open`.

homework/list.js
```diff
@@ -1,79 +1,85 @@
...
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+	preloadFirstNote : function () {
+		if (this._listNoteContent.length !== 0) {
```
snowmantw, line 32: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

snowmantw:
After check, I've found your content is almost the same with

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/pull/48

**I need you two to explain why your homeworks look like the same**
And you two better perform **really great** in the following lessons to avoid get into trouble.

PR #48, opened 2015-10-30
101502026

onlyfly34:
review=?@snowmantw

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,39 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function Content() {
+	this._wrapper = document.querySelector('#note-content-wrapper');
+}
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+Content.prototype = {
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+	start : function () {
+		window.addEventListener('note-open', (function(event) {
+		var note = event.detail;
+		this.resetWrapper();
+		this.drawNote(note);
+		}).bind(this));
+	},
+	resetWrapper : function () {
+		this._wrapper.innerHTML = '';
+	},
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+	drawNote : function (note) {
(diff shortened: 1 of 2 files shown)


homework/content.js
```diff
@@ -1,37 +1,39 @@
...
+'use strict';
+
+function Content() {
+	this._wrapper = document.querySelector('#note-content-wrapper');
+}
+
+Content.prototype = {
+
+	start : function () {
+		window.addEventListener('note-open', (function(event) {
```
snowmantw, line 10: Instead of using an anonymous function, set the listener as `this` and handle events with `handleEvent` method is better.

homework/content.js
```diff
@@ -1,37 +1,39 @@
...
+			  p.classList.add('note-passage');
+			  p.textContent = passage;
+			  buff.appendChild(p);
+		});
+		this._wrapper.appendChild(h);
+		this._wrapper.appendChild(buff);
+	  }
+};
+var content = new Content();
+document.addEventListener('DOMContentLoaded', function(event) {
```

homework/list.js
```diff
@@ -1,79 +1,86 @@
...
+'use strict';
+
+function List() {
+	this._listNoteContent = [];
+	this._wrapper = document.querySelector('#note-list-wrapper');
+}
+
+List.prototype = {
+	start : function () {
+		this.fetchList().then(function (data){
```
snowmantw, line 10: You can re-write the chain as

```
this.fetchList().then(this.updateList.bind(this))
                      .then(this.drawList.bind(this))
```

It's more concise.

homework/list.js
```diff
@@ -1,79 +1,86 @@
...
+			this.drawList.bind(this)
+		).then(
+			this.preloadFirstNote.bind(this)
+		);
+		window.addEventListener('click', function(event) {
+			this.onNoteOpen(event).bind(this);
+		});
+	},
+
+	onNoteOpen : function (event) {
```
snowmantw, line 23: Well it's a convention to prefix with `on` to handle the corresponding event, rather than to fire it. So, you should rename this function as `onClick` and to dispatch the `note-open`.

homework/list.js
```diff
@@ -1,79 +1,86 @@
...
+		if (event.target.classList.contains('note-title')) {
+			var id = event.target.dataset.noteId;
+			var content = this._listNoteContent[id];
+			window.dispatchEvent(new CustomEvent('note-open',
+			{ detail: content }));
+		};
+	},
+
+	preloadFirstNote : function () {
+		if (this._listNoteContent.length !== 0) {
```
snowmantw, line 33: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,79 +1,86 @@
...
+		  // Note: buff is captured, so we now have a
+		  // little closure naturally.
+			buff.appendChild(li);
+		});
+		ul.appendChild(buff);
+		this._wrapper.appendChild(ul);
+	},
+
+	fetchList : function () {
+		return new Promise(function (resolve, reject) {
```
snowmantw, line 63: Good, you have successfully "Promisified" the method.

snowmantw:
Where is your unit test??

snowmantw:
After check, I've found your content is almost the same with

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/pull/47

**I need you two to explain why your homeworks look like the same**
And you two better perform **really great** in the following lessons to avoid get into trouble.

PR #49, opened 2015-10-30
102502005

xxx663xxx:
## Tier 1 requirements met , including:
- Bind this in Event handlers
- Deal with object properties w/ constructor & prototype
- Deal with asynchronous flows with Promise
- Write my first test case for a pure function (as the last test of |test-list.js| file shows)

review=?@snowmantw

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,50 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
+var NoteContentManager = function() {
+	// Will be an element as the content wrapper.
+	this._wrapper = null;
+};
+
+NoteContentManager.prototype = {
+	
+	drawNote(note){
+		var title = note.title;
+		var h = document.createElement('h2');
+		h.textContent = title;
+		var passages = note.passages;
+		var buff = document.createDocumentFragment();
+		passages.forEach(function(passage) {
+			var p = document.createElement('p');
+			p.classList.add('note-passage');
+			p.textContent = passage;
+			buff.appendChild(p);
+		});
+		this._wrapper.appendChild(h);
+		this._wrapper.appendChild(buff);
+	},
+	
+	resetWrapper(){
+		this._wrapper.innerHTML = '';
+	},
+  
+	handleEvent(event) {
+		switch (event.type) {
+			case 'note-open':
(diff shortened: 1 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,50 @@
...
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
+  start() {
+    window.addEventListener('note-open', this);
```
snowmantw, line 43: It's good to see you bind the listener with the `this`, rather than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,91 @@
...
-  function onNoteOpen(event) {
-    if (event.target.classList.contains('note-title')) {
-      var id = event.target.dataset.noteId;
-      var content = _listNoteContent[id];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    };
-  }
+	fetchList(){
+		return new Promise((function(resolve, reject){
```
snowmantw, line 15: Good. You have successfully "Promisified" this method.

homework/list.js
```diff
@@ -1,79 +1,91 @@
...
+			// Note: buff is captured, so we now have a
+			// little closure naturally.
+			buff.appendChild(li);
+		});
+		ul.appendChild(buff);
+		this._wrapper.appendChild(ul);
+	},
+	
+	preloadFirstNote() {
+		if (this._listNoteContent.length !== 0) {
```
snowmantw, line 51: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,79 +1,91 @@
...
+  
+	handleEvent(event) {
+		switch (event.type) {
+		case 'click':
+			this.onNoteOpen(event);
+			break;
+		}
+	},
+  
+  	onNoteOpen(event) {
```
snowmantw, line 66: Well it's a convention to prefix with `on` to handle the corresponding event, rather than to fire it. So, you should rename this function as `onClick` and to dispatch the `note-open`.

homework/test/test-list.js
```diff
@@ -5,4 +5,9 @@ describe('Test > ', function() {
   it('will test some pure functions', function() {
     // Write any pure function assertion here.
   });
+  
+  it('simple test case', function(){
+	  assert.equal(1+1,2);
```
snowmantw, line 10: Please put this expression in one method of one manager, and call it here rather than just put the expression here.

PR #51, opened 2015-10-30
102502549

s102502549:
review=?@snowmantw

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,39 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function ContentManager() {
+    this._wrapper = document.querySelector('#note-content-wrapper');
+}
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+ContentManager.prototype = {
+    start: function () {
+        window.addEventListener('note-open', (function (event) {
+            var note = event.detail;
+            this.resetWrapper();
+            this.drawNote(note);
+        }).bind(this));
+    },
+    resetWrapper: function () {
+        this._wrapper.innerHTML = '';
+    },
+    drawNote: function (note) {
+        var title = note.title;
+        var h = document.createElement('h2');
+        h.textContent = title;
+        var passages = note.passages;
+        var buff = document.createDocumentFragment();
+        passages.forEach(function (passage) {
+            var p = document.createElement('p');
+            p.classList.add('note-passage');
+            p.textContent = passage;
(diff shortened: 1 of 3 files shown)


homework/content.js
```diff
@@ -1,37 +1,39 @@
...
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+ContentManager.prototype = {
+    start: function () {
+        window.addEventListener('note-open', (function (event) {
```
snowmantw, line 9: Instead of using an anonymous function, set the listener as this and handle events with handleEvent method is better.

homework/content.js
```diff
@@ -1,37 +1,39 @@
...
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+document.addEventListener('DOMContentLoaded', function (event) {
```

homework/list.js
```diff
@@ -1,79 +1,76 @@
...
+ListManager.prototype = {
+    start: function () {
+        this.fetchList().then(function (data) {
+            this.updateList(data);
+        }.bind(this)).then(this.drawList.bind(this)).then(this.preloadFirstNote.bind(this));
+        window.addEventListener('click', function (event) {
+            this.onNoteOpen(event);
+        }.bind(this));
+    },
+    onNoteOpen: function (event) {
```
snowmantw, line 17: Well it's a convention to prefix with on to handle the corresponding event, rather than to fire it. So, you should rename this function as onClick and to dispatch the note-open.

homework/list.js
```diff
@@ -1,79 +1,76 @@
...
+        if (event.target.classList.contains('note-title')) {
+            var id = event.target.dataset.noteId;
+            var content = this._listNoteContent[id];
+            window.dispatchEvent(new CustomEvent('note-open',
+                {detail: content}));
+        }
+    },
+    preloadFirstNote: function () {
+        if (this._listNoteContent.length !== 0) {
```
snowmantw, line 26: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,79 +1,76 @@
...
+            li.classList.add('note-title');
+            li.textContent = note.title;
+            // Note: buff is captured, so we now have a
+            // little closure naturally.
+            buff.appendChild(li);
+        });
+        ul.appendChild(buff);
+        this._wrapper.appendChild(ul);
+    },
+    fetchList: function () {
```
snowmantw, line 52: Good, you have successfully "Promisified" the method.

PR #53, opened 2015-10-30
102502038

leemiyinghao:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,39 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function ToDoItemReader() {
+  this._wrapper = document.querySelector('#note-content-wrapper');
+}
+ToDoItemReader.prototype = {
+    start: function() {
+	window.addEventListener('note-open', function(event) {
+	    var note = event.detail;
+	    this.resetWrapper();
+	    this.drawNote(note);
+	}.bind(this));
+    },
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+    resetWrapper: function() {
+	this._wrapper.innerHTML = '';
+    },
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+    drawNote: function(note) {
+	var title = note.title;
+	var h = document.createElement('h2');
+	h.textContent = title;
+	var passages = note.passages;
+	var buff = document.createDocumentFragment();
+	passages.forEach(function(passage) {
+	    var p = document.createElement('p');
+	    p.classList.add('note-passage');
+	    p.textContent = passage;
+	    buff.appendChild(p);
+	}.bind(this));
+	this._wrapper.appendChild(h);
+	this._wrapper.appendChild(buff);
+    }
+}
+document.addEventListener('DOMContentLoaded', function(event) {
+    var toDoItemReader = new ToDoItemReader();
+    toDoItemReader.start();
+});
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
--- a/homework/content.js
+++ b/homework/content.js
@@ -5,11 +5,7 @@ function ToDoItemReader() {
 }
 ToDoItemReader.prototype = {
     start: function() {
-	window.addEventListener('note-open', function(event) {
-	    var note = event.detail;
-	    this.resetWrapper();
-	    this.drawNote(note);
-	}.bind(this));
+	window.addEventListener('note-open', this);
     },
 
     resetWrapper: function() {
@@ -30,6 +26,12 @@ ToDoItemReader.prototype = {
 	}.bind(this));
 	this._wrapper.appendChild(h);
 	this._wrapper.appendChild(buff);
+    },
+
+    handleEvent: function(event) {
+	var note = event.detail;
+	this.resetWrapper();
+	this.drawNote(note);
     }
 }
 document.addEventListener('DOMContentLoaded', function(event) {
--- a/homework/list.js
+++ b/homework/list.js
@@ -1,79 +1,83 @@
 'use strict';
+function ToDoList(){
+    this._listNoteContent = [];
+    this._wrapper = document.querySelector('#note-list-wrapper');
+}
+ToDoList.prototype = {
+    start: function() {
+	this.fetchList(function(data) {
+	    this.updateList(data);
+	    this.drawList();
+	    this.preloadFirstNote();
+	}.bind(this));
+	window.addEventListener('click', function(event) {
+	    this.onNoteOpen(event);
+	}.bind(this));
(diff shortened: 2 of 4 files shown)


homework/content.js
```diff
@@ -1,37 +1,39 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+function ToDoItemReader() {
+  this._wrapper = document.querySelector('#note-content-wrapper');
+}
+ToDoItemReader.prototype = {
+    start: function() {
+	window.addEventListener('note-open', function(event) {
```
snowmantw: Instead of using an anonymous function, set the listener as `this` and handle events with `handleEvent` method is better.

homework/content.js
```diff
@@ -1,37 +1,39 @@
...
+	    var p = document.createElement('p');
+	    p.classList.add('note-passage');
+	    p.textContent = passage;
+	    buff.appendChild(p);
+	}.bind(this));
+	this._wrapper.appendChild(h);
+	this._wrapper.appendChild(buff);
+    }
+}
+document.addEventListener('DOMContentLoaded', function(event) {
```

homework/list.js
```diff
@@ -1,79 +1,83 @@
 'use strict';
+function ToDoList(){
+    this._listNoteContent = [];
+    this._wrapper = document.querySelector('#note-list-wrapper');
+}
+ToDoList.prototype = {
+    start: function() {
+	this.fetchList(function(data) {
```
snowmantw: Arrow function like `() => {}` can automatically bind the `this` for you.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+	    this.updateList(data);
+	    this.drawList();
+	    this.preloadFirstNote();
+	}.bind(this));
+	window.addEventListener('click', function(event) {
+	    this.onNoteOpen(event);
```
snowmantw: It's a convention to prefix method with `on` for the corresponding function, so it should be renamed as 'onClick'. And, to set `this` as the listener is better than an anonymous function.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+	    var content = this._listNoteContent[id];
+	    window.dispatchEvent(new CustomEvent('note-open',
+						 { detail: content }));
+	};
+    },
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+    preloadFirstNote: function() {
+	if (this._listNoteContent.length !== 0) {
```
snowmantw: nit: It's better to put the condition outside a function wholly wrapped by only one branch.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+		} else if (this.status !== 200 ){
+		    // Ignore error in this case.
+		    reject();
+		}
+	    };
+	    xhr.send();
+	}).then(function(){
+	    afterFetch(xhr.response);
+	}).catch(function(){
+	    console.log("fetchList failed.")
```
snowmantw: It's a serious issue in real code that to catch an Error but not re-throw it. Because the following chain can't now be interrupted by the error, so it could devastate your debugging strategy. If you want to log it, you should alway re-throw the error after that.

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+		    // The flow ends here.
+		    resolve();
+		} else if (this.status !== 200 ){
+		    // Ignore error in this case.
+		    reject();
+		}
+	    };
+	    xhr.send();
+	}).then(function(){
+	    afterFetch(xhr.response);
```
snowmantw: It's **wrong** here: if you are using a Promise, you **definitely** needn't a callback-style method. Instead of, you return a Promise. Please check Promise example like this:

 I suggest that you better to check and correct it, or you're just showing that you haven't understood Promise well.

homework/test/test-list.js
```diff
@@ -1,8 +1,35 @@
...
+	      },
+	      classList: {
+		  contains: function(input){
+		      return true;
+		  }
+	      }
+	  }
+      }
+      toDoList.onNoteOpen(onNoteOpenEvent);
+      assert.isNotNull(toDoItemReader._wrapper);
```
snowmantw: It's good to see you test a real method rather than a dummy pure method.

snowmantw:
After check, your critical error **look the same** as 101502012's:

https://github.com/MAP-NCU2015/Lesson-03-JavaScriptBestPratice-HW/pull/45

Please correct it **by yourself**.

leemiyinghao:
@snowmantw : I am not really understand what promise code should looked like, so I follow the code on page 93 of [slide](https://docs.google.com/presentation/d/1ofgD63SENpiihJ_MIYyZVW2TM11Mo-qwcstDbZytqIo/edit#slide=id.gbcb54ed09_1_1060).
I'm not sure if he (or she?) have the same ref as me.

So I need to correct the error on Github? Or explain on the class next week?

snowmantw:
You only need to correct it on this PR. And the page is not about the error: it is only a general reference for what the flow should look like.

The point is you implemented the Promise version of XHR method at line 59, but you still keep the callback style that I expect you to refactor. So you can keep only the Promise pattern, and remove the callback.

Also please note the `catch` in Promise will capture the error, and then the rest part of the chain will never know that. So in order to know the error early, you should print the error in the `catch` section, and then you definitely should re-throw that in the same section after you just print it out. If you don't re-throw that, even there is a critical error in the chain before the catch function, the rest part will still execute and it could do some unexpectedly behavior. This is a common mistake among your classmates' code.

homework/test/test-list.js
```diff
@@ -1,8 +1,36 @@
...
-  });
+      toDoItemReader.start();
+      toDoList._listNoteContent[0] = {
+	  title: "This is an pen",
+	  passages: [
+	      "This is a pencil",
+	      "That is a book"
+	  ]
+      };
+      var onNoteOpenEvent = {
```
leemiyinghao: Tried to create a real click event but fail.
Are there any method to set `event.target` to a fake `div` object? Or it's not needed?

homework/list.js
```diff
@@ -1,79 +1,83 @@
...
+		this.drawList();
+		if (this._listNoteContent.length !== 0){
+		    this.preloadFirstNote();
+		}
+	    });
+	promise.catch(
+	    (reason) => {
+		throw reason;
+	    }
+	);
```
leemiyinghao, line 21: Not sure if this is what Promise code should be like.
My understanding is pulling callback out from the function.

homework/test/test-list.js
```diff
@@ -1,8 +1,36 @@
...
+	      dataset: {
+		  noteId: 0,
+	      },
+	      classList: {
+		  contains: function(input){
+		      return true;
+		  }
+	      }
+	  }
+      }
```
leemiyinghao, line 31: Tried to create a real click event but fail.
Are there any method to set `event.target` to a fake `div` object? Or it's not needed?

homework/list.js
snowmantw, line 21: It's unfortunately wrong. You just got trapped by a common mistake. In fact, you should do this if you don't want to use the plain chaining style:

```
promise = promise.then(....);
promise = promise.catch(....);
```

Please note it's because every time the `then` or `catch` is called, it will create a new instance of `then-able`. So if you do this:

```
promise.then(task1)
promise.then(task2)
```

It will **NOT** guarantee `task2` is only after `task1`. In fact, they will execute **concurrently**. So you must make sure the `then-able` is updated as I wrote.

homework/test/test-list.js
snowmantw, line 31: Why did you think you need to create a real click event? I just mentioned that the name is wrong, because in the code it listen to `click` event, but handle it with the method named as on**NoteOpen**, not on**NoteClicked**.

snowmantw, line 31: And if you need to test a event in unit tests, you just create a plain object as:

```
{ type: 'click', details: ...}
```

And then feed it to the `handleEvent` method. You don't really need to create a click event.

PR #54, opened 2015-10-30
102502036

henryma68:


homework/content.js
```diff
@@ -1,37 +1,56 @@
...
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+(function (exports) {
+    var Contentmanager = function () {
```
snowmantw, line 4: Well CamelCase is a common convention.

homework/content.js
```diff
@@ -1,37 +1,56 @@
...
+
+
+        start() {
+            this._wrapper = document.querySelector('#note-content-wrapper');
+            window.addEventListener('note-open', this);
+
+        },
+
+        handleEvent(event) {
+            //promise
```
snowmantw, line 18: No?

homework/content.js
```diff
@@ -1,37 +1,56 @@
...
+            var buff = document.createDocumentFragment();
+            passages.forEach(function (passage) {
+                var p = document.createElement('p');
+                p.classList.add('note-passage');
+                p.textContent = passage;
+                buff.appendChild(p);
+            });
+            this._wrapper.appendChild(h);
+            this._wrapper.appendChild(buff);
+        },
```
snowmantw, line 49: nit: trailing comma

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
 
-  function preloadFirstNote() {
-    if (_listNoteContent.length !== 0) {
-      var content = _listNoteContent[0];
-      window.dispatchEvent(new CustomEvent('note-open',
-        { detail: content }));
-    }
-  }
+            this.fetchList()
+         .then((function (data) {
```
snowmantw, line 16: I recommend to use Arrow function like `() => {}` since it will bind the `this` automatically.

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
+             this.updateList(data);
+             this.drawList();
+             this.preloadFirstNote();
+         }).bind(this));
+            window.addEventListener('click', (function (event) {
```
snowmantw, line 21: It's better to create a `onClick` event to fire the `note-open` event, since it's a convention to add prefix `on` for the corresponding event.

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
+                this.onNoteOpen(event);
+            }).bind(this));
+        },
+
+        onNoteOpen(event) {
+            if (event.target.classList.contains('note-title')) {
```
snowmantw, line 27: I usually put the condition wraps the whole function outside the body. Name, putting it in the caller rather than the callee.

homework/list.js
```diff
@@ -1,79 +1,100 @@
...
+                var id = event.target.dataset.noteId;
+                var content = _listNoteContent[id];
+                window.dispatchEvent(new CustomEvent('note-open',
+                  { detail: content }));
+            };
+        },
+        handleEvent(event) {
```
snowmantw, line 34: In line 21 you `addEventListener` of click to an anonymous function, rather than to assign the `this` as the listener. So this method will do nothing and never be called.

snowmantw:
Please check how to use `addEventListener` with `this` and `handleEvent`. See:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#The_value_of_this_within_the_handler

PR #55, opened 2015-11-05
101403014 HW 

gmkevin00:
review?=@snowmantw
review?=@chungya
review?=@FlowerHop

Tier1 is finished.
I'm sorry about finishing this homework so late.

### DIFF
--- a/homework/content.js
+++ b/homework/content.js
@@ -1,37 +1,44 @@
 'use strict';
 
-(function() {
-  var _wrapper = document.querySelector('#note-content-wrapper');
+(function(exports) {
+	var NoteContentManager = function() {
+		this._wrapper = document.querySelector('#note-content-wrapper');
+	};
 
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+	NoteContentManager.prototype = {
+		start() {
+			window.addEventListener('note-open', function(event) {
+			  var note = event.detail;
+			  this.resetWrapper()
+			  .then((function(){
+					this.drawNote(note);
+			  }).bind(this));
+			}.bind(this));
+		},
+		
+		resetWrapper() {
+			  this._wrapper.innerHTML = '';
+			  return Promise.resolve();
+		},
 
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+		drawNote(note) {
+			  var title = note.title;
+			  var h = document.createElement('h2');
+			  h.textContent = title;
+			  var passages = note.passages;
+			  var buff = document.createDocumentFragment();
+			  passages.forEach(function(passage) {
+				var p = document.createElement('p');
+				p.classList.add('note-passage');
+				p.textContent = passage;
+				buff.appendChild(p);
+			  });
+			  this._wrapper.appendChild(h);
+			  this._wrapper.appendChild(buff);
+		}
+	
+	};
+	
+	exports.NoteContentManager = NoteContentManager;
 
-  function drawNote(note) {
-    var title = note.title;
-    var h = document.createElement('h2');
-    h.textContent = title;
-    var passages = note.passages;
-    var buff = document.createDocumentFragment();
-    passages.forEach(function(passage) {
-      var p = document.createElement('p');
-      p.classList.add('note-passage');
-      p.textContent = passage;
-      buff.appendChild(p);
-    });
-    _wrapper.appendChild(h);
-    _wrapper.appendChild(buff);
-  }
-
-  document.addEventListener('DOMContentLoaded', function(event) {
-    start();
-  });
-})();
+})(window);
--- a/homework/index.html
+++ b/homework/index.html
@@ -3,7 +3,8 @@
 <head>
   <meta charset="UTF-8">
   <title> Homework - Note List </title>
-  <script defer src="list.js"></script>
+  <script defer src="main.js"></script>
+  <script src="list.js"></script>
   <script defer src="content.js"></script>
   <link rel="stylesheet" type="text/css" href="main.css" />
 </head>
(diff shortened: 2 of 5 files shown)


homework/content.js
```diff
@@ -1,37 +1,44 @@
...
-  function start() {
-    window.addEventListener('note-open', function(event) {
-      var note = event.detail;
-      resetWrapper();
-      drawNote(note);
-    });
-  }
+	NoteContentManager.prototype = {
+		start() {
+			window.addEventListener('note-open', function(event) {
```
snowmantw, line 10: I prefer to use `addEventListener('note-open', this)` instead of binding on a function, since in this way all events could be redirected to `this.handleEvent`, and the `this` keyword in that method will refer to the instance you want to bind.

See this:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Example

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
+			  var note = event.detail;
+			  this.resetWrapper()
+			  .then((function(){
```
snowmantw, line 13: In fact, you can use Arrow function instead of binding every thing.
See:

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
-  function resetWrapper() {
-    _wrapper.innerHTML = '';
-  }
+		drawNote(note) {
+			  var title = note.title;
+			  var h = document.createElement('h2');
+			  h.textContent = title;
+			  var passages = note.passages;
+			  var buff = document.createDocumentFragment();
+			  passages.forEach(function(passage) {
```
snowmantw, line 30: Good to see the `forEach`.

homework/content.js
```diff
@@ -1,37 +1,44 @@
...
+			  this.resetWrapper()
+			  .then((function(){
+					this.drawNote(note);
+			  }).bind(this));
+			}.bind(this));
+		},
+		
+		resetWrapper() {
+			  this._wrapper.innerHTML = '';
+			  return Promise.resolve();
```
snowmantw, line 21: Well there is no need to use Promise here.

homework/index.html
```diff
@@ -3,7 +3,8 @@
 <head>
   <meta charset="UTF-8">
   <title> Homework - Note List </title>
-  <script defer src="list.js"></script>
+  <script defer src="main.js"></script>
+  <script src="list.js"></script>
```
snowmantw, line 7: It's recommended to use `defer` to avoid blocking page loading.

homework/list.js
```diff
@@ -1,79 +1,92 @@
...
+	var TodoListManager = function() {
+		this._listNoteContent = [];
+		this._wrapper = document.querySelector('#note-list-wrapper');
+	};
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+	TodoListManager.prototype = {
+		start() {
+			this.fetchList(function(data) {
```
snowmantw, line 11: Arrow function is better.

homework/list.js
```diff
@@ -1,79 +1,92 @@
...
+		this._wrapper = document.querySelector('#note-list-wrapper');
+	};
 
-  var _listNoteContent = [];
-  var _wrapper = document.querySelector('#note-list-wrapper');
+	TodoListManager.prototype = {
+		start() {
+			this.fetchList(function(data) {		
+				this.updateList(data)
+				.then((function(){
```
snowmantw, line 13: Use `.then(this.drawList.bind(this))` to improve the expressiveness.

homework/list.js
```diff
@@ -1,79 +1,92 @@
...
+					this.drawList(); 
+				}).bind(this)).then((function(){
+					this.preloadFirstNote();
+				}).bind(this)).catch((function(error){
```
snowmantw, line 17: Don't catch error and then silent it if you don't know what you're doing. Think about that: if an error really occur, you chain will eat it without any message. And the following chain will execute as usual. It could devastate your debugging strategy and cause lots of pain.

homework/list.js
```diff
@@ -1,79 +1,92 @@
...
+				}).bind(this));			
+			}.bind(this));
+			
+			
+			window.addEventListener('click', function(event) {
+				this.onNoteOpen(event);
+			}.bind(this));
+		},
+		 
+		onNoteOpen(event) {
```
snowmantw, line 27: Well it's a convention to prefix with `on` to handle the corresponding event, rather than to fire it. So, you should rename this function as `onClick` and to dispatch the `note-open`.

homework/list.js
snowmantw, line 17: And please be careful: this is a serious issue which definitely deserve a review- if you're working and set the review for this patch.

homework/list.js
```diff
@@ -1,79 +1,92 @@
...
+				window.dispatchEvent(new CustomEvent('note-open',{ detail: content }));
+			};
+		},
+		
+		preloadFirstNote() {
+			if (this._listNoteContent.length !== 0) {
+				var content = this._listNoteContent[0];
+				window.dispatchEvent(new CustomEvent('note-open',{ detail: content }));
+			}
+			return Promise.resolve();
```
snowmantw, line 40: Well even for the demo purpose you don't need to "Promisified" every method, although I know I once mentioned that we now have a final strategy to make every method as a Promise. However, the premise was you firstly know why you need that, so that you can explain the strategy for your reviewer if it's necessary.

homework/test/test-list.js
```diff
@@ -1,8 +1,24 @@
+var MathFunc=function(){};
+
+MathFunc.prototype  = {
```
snowmantw, line 3: It's better to register this method in the manager, rather than just write it in the test file.

snowmantw:
Overall you need to understand the details of Promise and event handler API more, and consider how to improve your expressiveness. For a homework to gain enough scores of course is the first priority, but you will probably get int trouble when you have a coding job. And that's basically what I try to teach all of you.

